PR #29, opened 2018-06-19
Added DbContext interfaces for EF configuration and operational store

xmichaelx:
Hello, this is a small but useful change. 

By adding  `IConfigurationDbContext` and `IPersistedGrantDbContext` interfaces we can reuse `AdminDbContext` when registering EF configuration and operational store in IdentityServer.

Then you should be able to add this context using following code:

```
services.AddIdentityServer()
                .AddConfigurationStore<AdminDbContext>()
                .AddOperationalStore<AdminDbContext>();
```

since both AddConfigurationStore & AddOperationalStore have constraint on T to be of type `DbContext` and implement `IConfigurationDbContext` or `IPersistedGrantDbContext`.

It has an additional advantage of making sure that all relevant DbSets are implemented by AdminDbContext.

HTH


skoruba:
Hello @xmichaelx ,

great idea, thank you!

PR #37, opened 2018-06-22
Added steps for using other database engines

xmichaelx:
Added steps for adding other database engines. With EF.Core it's pretty straightforward but still it's nice to have the steps for using other providers laid out.


skoruba:
@xmichaelx Great, I am happy for that. 👍 Thank you.

PR #39, opened 2018-06-28
Added autocomplete=off for secret values for scopes and clients

xmichaelx:
So that when typing in a new secret we don't get a list of previously typed in secrets.


skoruba:
Good! Thank you.

PR #41, opened 2018-07-08
Separation of the EF and Business Logic layers from the MVC

skoruba:


skoruba:
@xmichaelx Can you check it please?

xmichaelx:
Why you explicitly set table names in EF?

skoruba:
Good point. 👍  Do you recommend deleting it? Personally I like this - especially in Asp.Net Core Identity tables.
You can send your ideas please. :)

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/ComboBoxHelpers.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+using Skoruba.IdentityServer4.Admin.BusinessLogic.Dtos.Common;
+
+namespace Skoruba.IdentityServer4.Admin.BusinessLogic.Helpers
+{
+	public static class ComboBoxHelpers
+	{
+		public static void PopulateValuesToList(string values, List<string> list)
```
TomasHubelbauer: You should add a comment about what `values` or rename it to `jsonValues` or something, right now for the consumer it's not clear it's a JSON string.

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/ComboBoxHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+		{
+			if (string.IsNullOrEmpty(values)) return;
+
+			var listValues = JsonConvert.DeserializeObject<List<string>>(values);
+			if (listValues == null) return;
+
+			list.AddRange(listValues);
+		}
+
+	    public static void PopulateValue(string value)
```
TomasHubelbauer: Same here.

xmichaelx:
I didn't have issue with names provided by EF by default and for me it's just less code I have to maintain, but this is largely a personal preference, nothing big.

The other this is you have renamed some methods in IdentityController from UserSave and UserEdit and at least in VS this produces warnings:

```
Warning	CS0108	'IdentityController.User(string)' hides inherited member 'ControllerBase.User'. Use the new keyword if hiding was intended.
```

I recommend wrapping those User methods in block of:

```
#pragma warning disable CS0108
#pragma warning restore CS0108
```

Just so it's clear that we're doing this on purpose.

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/ComboBoxHelpers.cs
skoruba: Good idea, thank you @TomasHubelbauer 👍

skoruba:
@xmichaelx I've changed the methods back to `User` because previous names `UserSave` and `UserEdit` produce wrong `ViewName` - https://github.com/skoruba/IdentityServer4.Admin/blob/206685577296f99047a45aff413165b07080a617/src/Skoruba.IdentityServer4.Admin/ExceptionHandling/ControllerExceptionFilterAttribute.cs#L58-L68
Any idea how fix it? :)

xmichaelx:
@skoruba No ideas so far, I'll review changes tonight.

Since database access is nicely separated now, is connecting IS to use database that is being modified by Admin something we would contemplate instead of using InMemory implementations?

skoruba:
@xmichaelx Yes, absolutely, we can use the database instead of InMemory store.

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/ComboBoxHelpers.cs
skoruba: Done

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/ComboBoxHelpers.cs
skoruba: Done

skoruba:
@xmichaelx @TomasHubelbauer - Thank you guys for the review. :)

PR #45, opened 2018-07-25
Update seed data with default admin account

### LINKED ISSUE #44: How can I login?
It seems there is no default admin/admin user or something else like for login page.
Follow the step,
1. add-migrations
2. datebase update
3. uncomment seeddata

Then I start IDS4 at localhost:5000, and IDS4 admin at localhost:9000.
Where I open http://localhost:9000, the website redirect me to localhost:5000 login page.
But I don't have a username and password for login.
User table is empty when I check it in SQL Server Management Studio.
Is it something I have done wrong?

skoruba:
Fix #44


src/Skoruba.IdentityServer4.Admin/Configuration/Users.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Skoruba.IdentityServer4.Admin.Configuration
+{
+    public class Users
+    {
+        public const string AdminUserName = "bob";
+        public const string AdminPassword = "Pa$$word123";
+        public const string AdminEmail = "BobSmith@email.com";
```
TomasHubelbauer: Make sure this is not a real email or upon connection with an email server, someone will get spammed. :-) I suggest using @example.com or creating a site for the repo and a demo account inbox on it.

src/Skoruba.IdentityServer4.Admin/Helpers/DbMigrationHelpers.cs
```diff
@@ -31,12 +34,51 @@ public static void EnsureSeedData(IServiceProvider serviceProvider)
...
+        /// </summary>
+        private static void EnsureSeedIdentityData(UserManager<UserIdentity> userManager,
+            RoleManager<UserIdentityRole> roleManager)
+        {
+            // Create admin role
+            if (!roleManager.RoleExistsAsync(AuthorizationConsts.AdministrationRole).Result)
+            {
+                var role = new UserIdentityRole { Name = AuthorizationConsts.AdministrationRole };
+
+                roleManager.CreateAsync(role).Wait();
```
TomasHubelbauer: Why is this method not `async`? Used in a non-awaitable context?

src/Skoruba.IdentityServer4/Views/Account/Login.cshtml
```diff
@@ -17,7 +17,7 @@
 						<form asp-route="Login">
 							<input type="hidden" asp-for="ReturnUrl" />
 
-							<div class="alert alert-info" role="alert">Use username - <strong>bob</strong>, password - <strong>password</strong></div>
+							<div class="alert alert-info" role="alert">Use username - <strong>bob</strong>, password - <strong>Pa$$word123</strong></div>
```
TomasHubelbauer: Could this be pulled dynamically from the seed data? Just so that you don't have to update it in multiple places.

src/Skoruba.IdentityServer4.Admin/Helpers/DbMigrationHelpers.cs
skoruba: Great point! 
I am wondering about it as well. This method is executed in `Main` method from `Program.cs`. What do you recommend for this use case - try to rewrite it using async/await? I'm not sure how it will work. I've found similiar question on SO - https://stackoverflow.com/a/9212343/943403 - what do you think about it?

TomasHubelbauer: See the very last code block in that answer; `async Main` should work with the latest .NET Core.

You may need to have to set the language level to `Latest`:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#edit-the-csproj-file

skoruba: It works! :) 
Please take a look here - https://github.com/skoruba/IdentityServer4.Admin/blob/master/src/Skoruba.IdentityServer4.Admin/Program.cs#L10-L17 - this async method called `DbMigrationHelpers.EnsureSeedData(host)` is commented by default and if I add async keyword in `Main` then I get a warning about lacks of await keyword. Is there some potential side effect?

TomasHubelbauer: It should be fine, just change `public static void Main` to `public static async Task Main` and then you can `await` in it.

src/Skoruba.IdentityServer4.Admin/Program.cs
```diff
@@ -7,11 +8,11 @@ namespace Skoruba.IdentityServer4.Admin
 {
     public class Program
     {
-        public static void Main(string[] args)
+        public static async Task Main(string[] args)
         {
             var host = BuildWebHost(args);
 
-            //DbMigrationHelpers.EnsureSeedData(host);
+            //await DbMigrationHelpers.EnsureSeedData(host);
```
TomasHubelbauer: Instead of commenting it out you could add a build configuration which sets a variable and then `#if` it. This way it is not clear if commenting it out is intended.

Alternatively add a comment above this line saying when to comment out.

PR #49, opened 2018-07-31
Extend picker component with new functionality - show suggested items, add IdentityService tests

skoruba:
- The Picker provides new feature - the ability to show some suggested items (by default top 5 items):

image

- Replace some missing `@Html.Partial` with `async` version and correction of the namespaces.

- Add IdentityServer tests


src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/EnumerableExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    public static class EnumerableExtensions
+    {
+        public static IEnumerable<T> WhereIf<T>(this IEnumerable<T> query, bool condition, Func<T, bool> predicate)
+        {
+            return condition
+                ? query.Where(predicate)
+                : query;
+        }
+
+        public static IEnumerable<T> TakeIf<T, TKey>(this IEnumerable<T> query, Func<T, TKey> orderBy, bool condition, int limit, bool orderByDescending = true)
```
TomasHubelbauer: Could you replace `condition` and `limit` with `int?` instead? That way you have just one argument and the coupled arguments enforce correct invariants ("no value" or "some value" instead of current  yes/no value and the actual value which can encode invalid invariant where it says no limit but the number is non-zero).

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/EnumerableExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Skoruba.IdentityServer4.Admin.BusinessLogic.Helpers
+{
+    public static class EnumerableExtensions
+    {
+        public static IEnumerable<T> WhereIf<T>(this IEnumerable<T> query, bool condition, Func<T, bool> predicate)
```
TomasHubelbauer: Similarly here you could have `Func === null` be a replacement for `condition` and default it it `null`

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Helpers/QueryableExtensions.cs
```diff
@@ -13,6 +13,16 @@ public static IQueryable<T> WhereIf<T>(this IQueryable<T> query, bool condition,
                 : query;
         }
 
+        public static IQueryable<T> TakeIf<T, TKey>(this IQueryable<T> query, Expression<Func<T, TKey>> orderBy, bool condition, int limit, bool orderByDescending = true)
```
TomasHubelbauer: Same here

src/Skoruba.IdentityServer4.Admin.BusinessLogic/Repositories/ClientRepository.cs
```diff
@@ -55,21 +54,29 @@ public Task<Client> GetClientAsync(int clientId)
             return pagedList;
         }
 
-        public async Task<List<string>> GetScopesAsync(string scope)
+        public async Task<List<string>> GetScopesAsync(string scope, int limit = 0)
         {
-            var identityResources = await _dbContext.IdentityResources.Where(x => x.Name.Contains(scope)).Select(x => x.Name).ToListAsync();
-            var apiResources = await _dbContext.ApiResources.Where(x => x.Name.Contains(scope)).Select(x => x.Name).ToListAsync();
+            var identityResources = await _dbContext.IdentityResources
```
TomasHubelbauer: I see now how these extension methods are used so maybe your design makes  more sense seeing this call site. Still feels weird to me but at the moment I don't have a better suggestion.

src/Skoruba.IdentityServer4.Admin/Scripts/App/components/Picker.es5.js
```diff
@@ -100,6 +109,26 @@ ko.components.register('picker', {
...
+		this.getSuggestedItems = function () {
+
+			if (self.allowSuggestedItems() === false) return;
+
+			if (params.url) {
+				//start loading
+				self.loading(true);
+
+				//make ajax request and result add to search result
+				$.get(params.url + "=" + "&limit=" + self.topSuggestedItems(), function (data) {
```
TomasHubelbauer: Use `URLSearchParams` for this, not string concatenation

skoruba: I have used instead of string string concatenation - object as second parameter in get method, Thx. :-)

PR #53, opened 2018-08-12
Clean up PickerTagHelper, add missing userprofile's resource, add seed data via args param

skoruba:
Clean up PickerTagHelper - it is not necessary to use explicitly the html attributes for component props. Add seed data via args param. Add missing userprofile's resource. Add the username to the page header.


src/Skoruba.IdentityServer4.Admin/Program.cs
```diff
@@ -8,13 +9,24 @@ namespace Skoruba.IdentityServer4.Admin
 {
     public class Program
     {
+        private const string SeedArgs = "/seed";
+
         public static async Task Main(string[] args)
         {
+            var seed = args.Any(x => x == SeedArgs);
```
TomasHubelbauer: You can use `Contains` here.

src/Skoruba.IdentityServer4.Admin/Program.cs
```diff
@@ -8,13 +9,24 @@ namespace Skoruba.IdentityServer4.Admin
...
+            if (seed) args = args.Except(new[] { SeedArgs }).ToArray();
+
             var host = BuildWebHost(args);
 
             //NOTE: Uncomment the line below to use seed data
             //await DbMigrationHelpers.EnsureSeedData(host);
 
+            //Eventually use seed via command - dotnet run /seed
```
TomasHubelbauer: This comments says that eventually you will receive the arg, but to me it looks like this change makes it so that that's the case now. Is the comment still accurate?

src/Skoruba.IdentityServer4.Admin/Helpers/TagHelpers/PickerTagHelper.cs
```diff
@@ -78,12 +62,7 @@ private void AddComponent(TagHelperOutput output)
...
-                SearchResultTitle + "\", suggestedItemsTitle: \"" + SuggestedItemsTitle + "\", noItemSelectedTitle: \"" +
-                NoItemSelectedTitle + "\", searchInputPlaceholder: \"" + SearchInputPlaceholder + "\", showAllItemsTitle: \"" + ShowAllItemsTitle + "\", selectedItems: " + selectedItems + ", minSearchText: " + MinSearchText +
-                ", topSuggestedItems: " + TopSuggestedItems + " ,multipleSelect: " + MultipleSelect.ToString().ToLower() + " }}'></div>");
+            var rawPickerHtml = new HtmlString($"<div data-bind=\'component: {{ name: \"picker\", params: {{ search: \"\", hiddenId: \"{Id}\", url: \"{Url}\", selectedItemsTitle: \"{SelectedItemsTitle}\", allowSuggestedItems: {AllowSuggestedItems.ToString().ToLower()}, searchResultTitle: \"{SearchResultTitle}\", suggestedItemsTitle: \"{SuggestedItemsTitle}\", noItemSelectedTitle: \"{NoItemSelectedTitle}\", searchInputPlaceholder: \"{SearchInputPlaceholder}\", showAllItemsTitle: \"{ShowAllItemsTitle}\", selectedItems: {selectedItems}, minSearchText: {MinSearchText}, topSuggestedItems: {TopSuggestedItems} ,multipleSelect: {MultipleSelect.ToString().ToLower()} }}}}\'></div>");
```
TomasHubelbauer: You don't need to escape the `'` here around `data-bind` which will make the verbatim string a bit shorter.

TomasHubelbauer: Maybe something like this would work an would be more readable:

```cs
var component = new {
	name = "picker",
	@params = new {
		search = "",
		hiddenId = Id,
		url = Url,
		selectedItemsTitle = SelectedItemsTitle,
		allowSuggestedItems = AllowSuggestedItems.ToString().ToLower(),
		searchResultTitle = SearchResultTitle,
		suggestedItemsTitle = SuggestedItemsTitle,
		noItemSelectedTitle = NoItemSelectedTitle,
		searchInputPlaceholder = SearchInputPlaceholder,
		showAllItemsTitle = ShowAllItemsTitle,
		selectedItems = selectedItems, // Should this really be lowercase?
		minSearchText = MinSearchText,
		topSuggestedItems = TopSuggestedItems,
		multipleSelect = MultipleSelect.ToString().ToLower()
	}
};
$"<div data-bind='component: {JsonConvert.SerializeObject(component)}'>"
```

src/Skoruba.IdentityServer4.Admin/Program.cs
skoruba: I meant for seeding you can uncomment line 22 or use command `dotnet run /seed`. 
You are right it isn't probably clear. Can you help me with better comment? Maybe remove lines 21,22 and use only seeding via args?

TomasHubelbauer: I see what you mean now, so maybe this would work as a replacement for 21-28:

```cs
// Uncomment this to seed upon startup, alternatively pass in `dotnet run /seed` to seed using CLI
//await DbMigrationHelpers.EnsureSeedData(host);
if (seed)
{
    await DbMigrationHelpers.EnsureSeedData(host);
}
```

This way it's colocated and makes more sense to me, in my original comment I didn't notice the two blocks of code were related before taking another look after your comment.

skoruba: It is great, thanks! :-)

PR #91, opened 2018-10-08
Added check for typically forbidden whitespaces in the username

xmichaelx:
More often than not users try to put their Name Surname into Username field so this helps avoiding getting and exception thrown into their face.


skoruba:
Works fine, thanks!

PR #93, opened 2018-10-17
Add gravatar support to users/usersprofile

Sindrenj:
Adding a Gravatar/image to each user will make it easier to identify users.
This request will add a Gravatar-image to each user in the users-list. It will also add an image to the user-profile.


skoruba:
Looks great! :)
Thanks @Sindrenj 👍

PR #96, opened 2018-10-19
 Update missing Simplified Chinese localization

Seaear:
Update missing Simplified Chinese localization


skoruba:
Great, thanks!

PR #110, opened 2018-11-06
 Optimize Simplified Chinese translation

Seaear:


Seaear:
Optimize simplified Chinese translation and Add Language Select

skoruba:
Please, take a look - the resources are stored twice. Can you fix it? Or may I do it manually later?
Thanks.

skoruba:
BTW: What do you think about issue with lowercase language code as a suffix of file? Some guy mentioned that had some issue on Linux with it?
Thx.

skoruba:
OK, thank you!
I will check it and merge it.

PR #130, opened 2018-12-15
Add russian translation, polish chinese translation

skoruba:
Add russian translation, polish chinese translation, add select for traslatation.

@Seaear @alev7ina - Thank you for your great work! Could you please check it?


skoruba:
This PR will be part of update to beta5.

PR #147, opened 2019-01-15
Make AdminDbContext compatible to mysql

h3rr-d:
Remove HasColumnType("xml"). It's mapped to a string anyway.


xmichaelx:
@svendu I'll check it out on Ubuntu and MySQL probably at the weekend and I'll approve it.

PR #151, opened 2019-01-21
Add tutorial for azure deploy

xmichaelx:
Added tutorial for deploying to Azure


docs/Configure-Azure-Deploy.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Create database
+
+If you don't have publicly accessible database you will need to create one. Follow tutorials for creating databases on Azure:
+
+- [SQL Server](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-get-started-portal)
+- [PostgreSQL](https://docs.microsoft.com/en-us/azure/postgresql/quickstart-create-server-database-portal)
+- [MySQL](https://docs.microsoft.com/en-us/azure/mysql/quickstart-create-mysql-server-database-using-azure-portal)
+
+Replace connection strings in appsettings with connection string to generated databse.
```
TomasHubelbauer: Typo in `databse`, also would change `appsettings` to `appSettings.json`

docs/Configure-Azure-Deploy.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Deploying webbaps to Azure App Service
+
+We will assume in the tutorial that STS and Admin were deployed to:
+
+- https://is4-sts.azurewebsites.net - STS
+- https://is4-admin.azurewebsites.net - Admin panel
+
+
+### Updating urls
```
TomasHubelbauer: I would capitalize URL here

docs/Configure-Azure-Deploy.md
```diff
@@ -0,0 +1,77 @@
...
+	"IdentityServerBaseUrl": "https://is4-sts.azurewebsites.net"
+}
+```
+
+Then follow instructions from [Quickstart deploy to Azure](https://docs.microsoft.com/en-us/visualstudio/deployment/quickstart-deploy-to-azure) 
+
+
+### Adding certificate for signing tokens
+
+We also need to upload pfx certificate for signing tokens. If you don't have one here are the steps to do it using openssl (remember to write down the password - we will need it later):
```
TomasHubelbauer: It would be good to link OpenSSL installation instructions on Windows. Or an alternative Windows OOTB way of generating these certificates.

xmichaelx:
Added suggestions from Tomas, I think that using openssl for certificate generation both on windows & linux is a good idea.

Issue #139, opened 2019-01-06
Allow for email to be used instead of username in the logging screen


xmichaelx:
Some websites allow for this in order to have one string less to remember. It would be nice to consider it as an option.

skoruba:
Done on `master`. Please check new release.
Thanks!

chungonion:
Hi there, just to my curiosity, is that possible for user to enter either email / username at the same time? Thanks!

PR #152, opened 2019-01-24
Allowing for using email in addition to username

xmichaelx:
If user will enter instead of username his email this will check if such email exist and then proceed with authentication. 

In order to prevent from guessing username from email whatever user enters will be always returned back to him (thus usage of additional local variable userName).

We can think about adding info that email can be also supplied into localization.

This provides functionality mentioned in #139


xmichaelx:
Added option for configuring desired behavior for login resolution:
- use username
- user email
- use username or email (if input is a valid email)

In configuration there is additional entry:

```json
"LoginConfiguration": {
  "ResolutionPolicy": "Username"
}
```

xmichaelx:
Updated for case where someone has usernames that are valid emails.

PR #161, opened 2019-02-03
Add profile management to STS, change int to default guid for Identity

skoruba:
- Add profile management - Account linking, Password reset, Two-Factor Authentication (2FA)
- Add support for Sendgrid
- Change int to default guid for Identity
- Add editing mode for Picker component


skoruba:
@duki994 - Thank you for good points!

duki994:
@xmichaelx 
> Some issue with cookies not being cleaned up after password reset but other than that everything seems to work - we can check that issue out on dev branch.

You have to use SignInManager SignOutAsync to sign out async.

HttpContext SignOutAsync should sign-out all authentication schemes.

**BUT (a big but)**
ASP.NET Identity uses internally it's own authentication scheme so HttpContext.SignOutAsync won't work (weird and wrong from my point of view, but that's how it is).

That's a *little* reason why SignInManager and SignOutAsync actually exist in ASP.NET Identity.

So you need to call explicitly SignInManager's SignOutAsync.


dotPeek and many IL decompilers have witnessed how much time many of my colleagues and I lost on this issue. 
So this should be documented too save innocent developer from inferno of multiple auth schemes and multiple "same" implementations :)

Found the issue while writing comment: https://github.com/IdentityServer/IdentityServer4/issues/2087


I will do a PR on weekend to add this stuff to docs. 
Simple, yet very much time saving for someone using and/or modifying this solution.

skoruba:
@duki994 - We are using `await _signInManager.SignOutAsync();`

PR #175, opened 2019-02-18
fixed user claim add form

baywet:
When adding a user claim I was always getting the error "Claim Type is required".
This is because the deserilisation of Model.ClaimType is not working properly, resulting in a non-functioning multi-select control.
Multi-select is not required in this case anyway. So I replaced it by a text field instead

### DIFF
--- a/templates/template-publish/content/src/SkorubaIdentityServer4Admin.Admin/Views/Identity/UserClaims.cshtml
+++ b/templates/template-publish/content/src/SkorubaIdentityServer4Admin.Admin/Views/Identity/UserClaims.cshtml
@@ -40,18 +40,7 @@
 							@await Html.PartialAsync("User/Section/Label", "Type")
 						</label>
 						<div class="col-sm-9">
-							<picker id="ClaimType" multiple-select="false" min-search-text="2"
-							        selected-item="@Model.ClaimType" url="@Url.Action("SearchClaims", "Configuration")?claim"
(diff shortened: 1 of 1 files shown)


skoruba:
@baywet - Did you try to click on item in search result in picker component? In this component is set parameter to `multiple-select="false"` - it should work correctly.

baywet:
I type in my claim and then click on the blue box under it. I'm using chrome.

senseicz:
:(

skoruba:
I will test it, thank you.

senseicz:
Apologies for sending my previous email, combination of fat fingers and
smart watch with reply functionality.

I'm sorry :)

skoruba:
It looks OK, currently tested on Chrome, IE - take a look at:

image

Correctly saved as well.

baywet:
something might be misconfigured on my end, I don't get any suggested items and when I click on the blue button (with the claim name) after typing, I doesn't become white, and the crayon and cross icons don't appear next to it. Any idea why it'd behave like that in my environment?

skoruba:
Please take a look at registering of taghelpers in your solution. Here: src/Skoruba.IdentityServer4.Admin/Views/_ViewImports.cshtml

skoruba:
Please, if you will have still problem with picker component, feel free open issue.
I'll close this PR.
Thank you.

PR #180, opened 2019-02-20
Optimize Chinese translation

Seaear:


skoruba:
Please, can you translate these two new labels:

> Return back to IdentityServer
> Authorization Code Flow with PKCE

Thank you :)

PR #182, opened 2019-02-20
Small reorganization of STS page so it looks more like Admin panel

xmichaelx:
Layout proposed in the PR:

obraz

basically change is so it looks like the admin panel instead of hiding it in the dropdown in the top right corner


src/Skoruba.IdentityServer4.STS.Identity/Views/Shared/_Layout.cshtml
```diff
@@ -52,15 +52,10 @@
                     @Localizer["Settings"] - @name
                 </button>
                 <div class="dropdown-menu">
-                    <a class="dropdown-item" asp-controller="Manage" asp-action="Index">@Localizer["ManageProfile"]</a>
```
skoruba: Please, don't remove these menu items.

Issue #187, opened 2019-02-21
Expect to add the Exceptionless log client support


Seaear:
https://github.com/exceptionless/Exceptionless

> The definition of the word exceptionless is: to be without exception. Exceptionless provides real-time error reporting for your JavaScript, Node, .NET Core, ASP.NET, Web API, WebForms, WPF, Console, and MVC apps. It organizes the gathered information into simple actionable data that will help your app become exceptionless!

skoruba:
#268 - will be solved here.

PR #191, opened 2019-02-22
Add Exceptionless client

Seaear:
#187


skoruba:
Thanks for this PR, let me check the Exceptionless - it seems like paid product, right?

Seaear:
Exceptionless is free to use and can be privatized for deployment

skoruba:
I will add this configuration for Exceptionless into our docs.
If somebody will want to add exceptionless - it will be available from our docs.

Thanks!

PR #213, opened 2019-03-07
Move client id and scopes to application.json for changeable after deploy.

PlanCZero:
As a developer, I would like to be able to change client id and scopes after deploy admin without re-build and re-deploy.


skoruba:
Great idea! :)

PR #214, opened 2019-03-07
fix translations fa resources

saeedrahmo:
persian fix


skoruba:
Thanks! 👍 

Can you please translate these two labels:

https://github.com/skoruba/IdentityServer4.Admin/blob/dev/src/Skoruba.IdentityServer4.Admin/Resources/Views/Configuration/Client/Section/Name.fa.resx#L124

https://github.com/skoruba/IdentityServer4.Admin/blob/dev/src/Skoruba.IdentityServer4.Admin/Resources/Views/Account/AccessDenied.fa.resx#L127

saeedrahmo:
yes already don.

PR #218, opened 2019-03-11
Grants should be visible for authenticated user

xmichaelx:
I know that in IS4 samples it was one of the ways to login, but since we have login - we can use it :)

And it doesn't look bad:
obraz


xmichaelx:
Now even better with login being the first panel visible (assuming you read from left to right)

obraz

skoruba:
Thanks! 👍

Issue #55, opened 2018-08-15
Use in existing IdentityServer setup


ruant:
For people having identity server running already, any plans on making a guide on how to get this awesome Admin UI to work?

Maybe there has to be some work done with exposing setup options like naming, using existing dbcontexts etc.

ruant:
Especially if it's planned to release this UI as a simple injectable middleware as issue #28 is about.

skoruba:
@ruant Sure, I will add this use case to the upcoming video tutorial. It's a great idea. 👍 

We are currently working on a project template for dotnet cli. It will be possible to use like this:
`dotnet new skoruba.is4admin`. #51
I think - it might be useful for the better customization. You can define your administration name, default admin role name etc.

After that I want to create the nuget packages with UI and BL & EF layers, but I think that takes more time to complete. :)

Thank you for your feedback.

Brice-xCIT:
Hi! This issue to me should be priority number one. After spending several hours in the code I still don't see how to use this with an existing installation. Somehow the code forces consumers to use its own EF migration and database schema, which bundles ASP.NET Core Identity and IdentityServer4.EntityFramework models into one single DB and DbContext. I tried to decouple both of these things to bind them to two different EF connections, without any success. Similarly, using this single DB as operational store, config store and identity store in vanilla IS4 fails because (from what I understand) migrations are not compatible. 

I would really love some doc about how to use this nice admin UI with the plain old IS4 quickstart at http://docs.identityserver.io/en/release/quickstarts/8_entity_framework.html.
At least, is it even possible?

skoruba:
Hi,
I absolutely understand your requirements, I will create a documentation for that.
Please for now - take a look at this branch - ´feature/extensible-aspnet-identity´ -
https://github.com/skoruba/IdentityServer4.Admin/tree/feature/extensible-aspnet-identity
Currently, I am working on better using of this administration with the existing instance of the IS4.
Do you want to use the administration only for IS4 or with Asp.Net Core Identity?
Thanks!

Brice-xCIT:
Thanks for your message and responsiveness, @skoruba!
I found out about the branch right after writing the message, and although it definitely goes into the right direction, as far as I can see AdminDbContext still wraps the three stores.
[...]

Issue #28, opened 2018-06-04
encapsulate project to injectable part


snys98:
I'm wondering if it is possible to change the structure of this project to two different project(nuget package).

- **IdentityServer4.Admin**
this project will register all operations to a specific endpoint(like "/identity-server-admin") as a rest api.

- **IdentityServer4.Admin.UI**
this project will register all default page to a specific endpoint(like "/identity-server-admin-ui") and it would be possible to inject some css or js or something.

Just like what swagger&swagger-ui has done.

Besides, thanks for your huge efforts for this repo.

skoruba:
Great idea @snys98.

I think it's absolutely possible. I want to prepare the REST api like you mentioned.

What do you suggest with UI part?

Thank you for your feedback. :)

snys98:
I used to modify [something](https://github.com/microexs/Microex.Swagger) of injecting a UI. 

Here register the middleware which serves the index.html and enable directory browse into the assembly(where we can put pages there):
https://github.com/microexs/Microex.Swagger/blob/8159f6ac84c9672ea938a0f1d760035436641e62/Microex.Swagger/SwaggerUI/SwaggerUIBuilderExtensions.cs#L19

And here to inject things we need to customize the pages.
https://github.com/microexs/Microex.Swagger/blob/8159f6ac84c9672ea938a0f1d760035436641e62/Microex.Swagger/SwaggerUI/SwaggerUIIndexMiddleware.cs#L47
https://github.com/microexs/Microex.Swagger/blob/8159f6ac84c9672ea938a0f1d760035436641e62/Microex.Swagger/SwaggerUI/index.html#L73

But considering there are multiple pages and all these pages requires authentication of a role like "IdentityServerAdmin", it would be much more harder to achieve this.

skoruba:
I am working on REST api, but I am wondering how distribute the set of controllers as nuget package - what do you think about it @snys98? 

I think - it is possible to add into nuget package some repositories and services that are connected with database model but what about api endpoints? :)

snys98:
Maybe we should forget about the MVC pattern but try to implement a middleware instead.

all these could be done with a single line in the pipeline, like:
```
app.UseIdentityAdmin(options=>{
        option.EndPoint = "identity-admin";
        // more configs...
}
```
Then api endpoint("/identity-admin") should deal with all the request that has the url prefix of  "identity-admin" then map the sub-urls to corresponding handler functions.

skoruba:
[...]

PR #225, opened 2019-03-19
Extracted core library to use in existing setups

Brice-xCIT:
This pull request provides the ability to **target Skoruba.IdentityServer4.Admin as a netstandard2.0 library**. It also adds IServiceCollection and IApplicationBuilder façade extensions as well as an associated public options model in order to **configure the app programmatically** rather than using appsettings.json. 

I took every possible effort to retain compatibility with other use cases that were previously supported (configuration read from appsettings.json) but further testing may be required.

Overall, I hope these changes will make it easier for people like me to use this great library in existing setups. Would Skoruba.IdentityServer4.Admin ship as a NuGet package, it would be now directly usable in an Asp.Net Core app in a similar fashion to how IdentityServer4 itself is used.

This PR helps fixing #55 and #28 as far as I'm concerned, but more testing is probably needed.

### Notable changes

- The privileged way to use the library is now to use `services.AddIdentityServerAdminUI(...)` and `app.UseIdentityServerAdminUI()`. 
Methods in StartupHelper, although still public, are only used internally by these extension methods, and by test projects.

- Configuration has been refactored to support programmatic options being passed to `services.AddIdentityServerAdminUI(options => ...)`. 
`IConfigurationRoot` instances are no longer used internally, but can still be imported using `IdentityServerAdminOptions.ApplyConfiguration(IConfigurationRoot)`

- The Asp.NET Core app previously called Skoruba.IdentityServer4.Admin is now called Skoruba.IdentityServer4.Admin.App and uses the same newly introduced extension methods.

- Configuration interfaces have been removed because they were never really used anyway.

- All internal code referencing single DbContexts has been removed, because unused.


Brice-xCIT:
Aaaaaand it doesn't pass tests. Woops...

Brice-xCIT:
Pfew, all sorted out now 😅

skoruba:
Thank you @Brice-xCIT 👍 
I'll check it. :)

skoruba:
BTW: It is amazing 👍

skoruba:
@Brice-xCIT - What about the use case - if you want to rewrite some View/Controller from `Skoruba.IdentityServer4.Admin`?
I tested the creation of HomeController in `Skoruba.IdentityServer4.Admin.App` - but it is not possible to rewrite the original HomeController from `Skoruba.IdentityServer4.Admin` - any idea? :)
Thanks!

Brice-xCIT:
Hmm. In what case would you want to do that?
Intuitively, this would call for extensibility of the base controllers, and the ability to add custom controllers in the options. Would that be only for the home controller or all controllers?

skoruba:
Maybe I want to use my own implementation of `Users` view and some extra authorization stuff, so I need access to Controller as well. I want to be able to rewrite `Users.cshtml` and IdentityController.
Any idea?
Thanks.

Brice-xCIT:
I think that, for such a big amount of changes and customization, a user would have to download the library code and modify it directly. Which is exactly what they would have had to do before this PR anyway, isn't it? :)

skoruba:
Yes, I am wondering about [razor class library](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-2.2&tabs=visual-studio) - move these views, controllers into Areas.
Currently is possible to install this project via dotnet new template - where you will have a UI (Views/Controllers/Resources) served in project after installation. :)
But you are right, everything is about compromises, because if you want to create massive changes, it is better download whole solution.
BTW: What do you think about razor class library above?

Brice-xCIT:
I haven't used razor class libraries yet, but it seems really nice and probably the right solution if you want the consuming app to be able to override the pages.

Regarding the dotnet new template command, is the use case again to be able to customize the pages? It seems kind of weird to me that a consuming app would want to change anything more than pure aesthetics, without cloning the source code and modifying it there. 

This said, it does seem like RCLs may be the best way to support this use case as well. However, I'm not sure how much of an effort it would be to refactor the core library that way.

skoruba:
Thank you @Brice-xCIT - I will continue on this branch with extracting UI as nuget package using `Razor Class Library`.

Brice-xCIT:
Awesome, thanks!

sw-ms-pradipbaria:
@skoruba, first Big Thanks for this project. 
I am looking for extracting UI as nuget package. Can i have approx timeline when it will be available? It will be grate if we use is as package.

skoruba:
@sw-ms-pradipbaria - Thanks for your request - currently I am working on adding of new API. Separate UI package is in process. @Brice-xCIT did a lot of work on separation of UI into package and I will definetely continue on this, but later. :)

sw-ms-pradipbaria:
Thanks for updates @skoruba , keep eyes on PRs.

PR #229, opened 2019-03-27
Append 'Async' to name of async methods

marcelstuder:
Some methods that return a Task<T> were missing the 'Async' part in their name (e.g. GetPersistedGrantByUser). 

They have been renamed to GetPersistedGrantByUser**Async** and the like.


skoruba:
Thanks!

PR #231, opened 2019-03-27
Added configuration from json file

xmichaelx:
Configuration read from json file. What do you think? I'd like to remove hardcoded code in the next commit :)

Client secrets are hashed before being pumped into db.


skoruba:
Great work, I will test it. 👍
Thank you

xmichaelx:
I'll add additional code for deserializing System.Security.Claims from Json, because by default they have trouble with that. So do not pull this just yet.

Issue #183, opened 2019-02-20
Upon login keeps redirecting back to login page.


MaryadaPurshottam:
I just downloaded this. Ran migrations and seeddata and then ran both projects. But when i login using the existing username and password stored wtihin Identity.Users.cs, it keep redirecting me back to the login page. I download it and ran the migrations as well as the seeddata.  this is the returnurl

"/connect/authorize/callback?client_id=skoruba_identity_admin&redirect_uri=http%3A%2F%2Flocalhost%3A9000%2Fsignin-oidc&response_type=id_token&scope=openid%20profile%20email%20roles&response_mode=form_post&nonce=636862813423997478.NDA0MmRkYzQtODdiMC00NWYzLTk0MGUtZTI5NDkzNjMyOTcyZjVhNWI2OWEtYjc5OC00OGJhLTk0OWYtMzY4MzE3MmE1Mjg2&state=CfDJ8DaiEcgA_YNIijFvat1kQrzHyDUpK0JXm-wTSG4MyrmgBP_rqyL6UpKmJ_vnu0boipJgILPRnS6hMaCWxnc9DR2EPtcJLgdNPNdgyWQRwXDS6CgCiMGJmJb74wC2-rKLzf-ny7_IwXdl-FTccTqkNQLpj--6G8NPAXO8UOYLbflkJrkNkH0Du4dDwKHZJmIPzjWa402F1ecyszJ1T4MqFyKhP0ZtWeSwlo6pG03IjVFWBOzp4JDaXKl7ieuMllZQtn56I4iPgnLhxviE6JiJ2Mk8vqVYQAbWZfB5qE82LWCWi8Bde3P8WonMcQ-2w2315g&x-client-SKU=ID_NETSTANDARD2_0&x-client-ver=5.3.0.0"

_Originally posted by @MaryadaPurshottam in https://github.com/skoruba/IdentityServer4.Admin/issues/44#issuecomment-465685742_

skoruba:
Can you hit breakpoint here: https://github.com/skoruba/IdentityServer4.Admin/blob/master/src/Skoruba.IdentityServer4.STS.Identity/Quickstart/Account/AccountController.cs#L109
- is `result` succeeded? 

Please, send more details.
Thank you.

MaryadaPurshottam:
Yep, result.Succeeded comes out true. It then proceeds to line 124 to redirect with to the above url which shows the login page once again.

MaryadaPurshottam:
[...]

PR #246, opened 2019-04-07
Clear out IdentityServer cookies on logout.

duki994:
This will prevent weird infinite loop redirects that may happen.

#183 will be fixed by this.


skoruba:
I think Asp.Net Core Identity by default uses another cookie, which we are clearing during logout process.
It comes from sample code from IdentityServer4.
Why clear this cookie?
Thanks!

skoruba:
Closing because of comment above.
Feel free to send some additional info about that.
Thanks!

PR #249, opened 2019-04-12
updated logic for return url

martea:


martea:
@YouthLab was thinking of defaulting the ReturnUrl to "/" then we dont need to if cases

skoruba:
@martea - I agree with @YouthLab - this refactoring will be fine. ;)
Thanks!

martea:
adjusted the pr :)

Issue #253, opened 2019-04-18
Add Swedish (sv) localization


Zyxious:
I've started working on Swedish localization.
Can make a PR when I'm done.

skoruba:
That is great, thanks!

skoruba:
Great job, thanks! 👍

Zyxious:
You're welcome! 
Thanks for a great project!

PR #254, opened 2019-04-18
#253 Added Swedish (sv) localization

Zyxious:
Added Swedish (sv) localization.


skoruba:
Great work, let me check it 👍 
Thanks!

PR #255, opened 2019-04-18
Simplified applying configuration in options

Brice-xCIT:
Refactored options ApplyConfiguration method to take IConfiguration instead of IConfigurationRoot in order to be easier to use with DI in Startup (IConfiguration can be DI'd, not IConfigurationRoot).


skoruba:
Thanks!

PR #263, opened 2019-04-26
Converted Admin project to Razor Class Library

Brice-xCIT:
This PR converts the project type of Skoruba.IdentityServer4.Admin from netstandard 2 library to Razor Class Library, enabling in theory to override views in consuming applications and to overall facilitate deployment. 

Views have not been ported to Areas, so ```.WithRazorPagesRoot("/Views")``` is used to break away from the Razor default name "Pages".


skoruba:
Great job, I will test it.
Thanks!

PR #285, opened 2019-05-26
auto generate strong secret key

mperk:


skoruba:
Sorry for delay - can you please move all javascript code into separate file? :)
Thanks!

skoruba:
Thank you @pbros - I will move your js code into separate JS file, great idea! 👍

PR #290, opened 2019-06-01
Docker withouth /etc/hosts modification

xmichaelx:
I'm  tagging @Seaear because I've largely based this on his work. @Seaear if you want I can create a PR to your repo and then we can pull your older PR in.

We have two services:

- Skoruba.IdentityServer4.STS.Identity with `sts` hostname assigned inside default bridge network controller mapped to `localhost:5000`

- Skoruba.IdentityServer4.Admin with `admin` hostname ssigned inside default bridge network controller mapped to `localhost:5001`

But unfortunately Admin cannot perform requests to `localhost:5000` only to `sts:5000` so fine, let's put that in Admin configuration, but then what? 

Discovery endpoint generates everything prepended with `sts:5000` instead of `localhost:5000` including authorize endpoint to which we're redirected to type in our credentials and since we're not inside the internal network between the host we get nothing.

What was the solution? To use option added to IdentityServer to change the IssuerUri and use that changed IssuerUri to manually set the the base url for `authorize_endpoint` and `issuer` so it doesn't depend on who is asking. And this solved it. Everything works without any modifications to your /etc/hosts.

tl;dr modifying discovery document so this is what admin services sees within the docker network solved the issue:

```
{
  "issuer": "http://localhost:5000",
  "jwks_uri": "http://sts:5000/.well-known/openid-configuration/jwks",
  "authorization_endpoint": "http://localhost:5000/connect/authorize",
  "token_endpoint": "http://sts:5000/connect/token",
  "userinfo_endpoint": "http://sts:5000/connect/userinfo",
  "end_session_endpoint": "http://sts:5000/connect/endsession",
  "check_session_iframe": "http://sts:5000/connect/checksession",
  "revocation_endpoint": "http://sts:5000/connect/revocation",
  "introspection_endpoint": "http://sts:5000/connect/introspect",
  // ... the rest
}
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,59 @@
+version: '3'
+
+services:
+  admin:
+    build:
+      context: .
+      dockerfile: src/Skoruba.IdentityServer4.Admin/Dockerfile
+    ports:
+      - "5001:5001"
+    networks:
+      - host
+    environment:
+      - ASPNETCORE_ENVIRONMENT=Development
+      - ASPNETCORE_URLS=http://+:5001
+      - "ConnectionStrings:ConfigurationDbConnection=Server=db;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true"
+      - "ConnectionStrings:PersistedGrantDbConnection=Server=db;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true"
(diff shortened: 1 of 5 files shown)


xmichaelx:
Added issue on IdentityServer4 side: IdentityServer/IdentityServer4#3307

Maybe wise people there will be able to suggest more elegant and maintainable solution.

skoruba:
@xmichaelx  - Thanks for this PR, we will continue with docker support in PR - #367, but I don't want to remove this branch yet.

PR #303, opened 2019-06-27
Fixed writing accesses that bypass manager classes

gingters:
In our custom IdentityServer implementation, we need to create events for every change on the users and roles.

For that, we decorated the `UserStore<TUser>` of the AspNetCore Identity model and create our events on every write access. The AspNetCore.Identities `UserManager` and `RoleManager` both use the `IUserStore` and `IRoleStore` for write access exclusively, so that is the only good place to intercept changes to the users and roles.

In your `IIdentityRepository` you use the manager classes to _create_ user claims and role claims, but for deleting these you bypass the `UserManager` / `RoleManager` classes and use the `IdentityDbContext` directly to delete `UserClaims` and `RoleClaims`.

This pull request changes this to also use the manager classes (and thus internally the Store) to write these changes to the database, so that we are capable of intercepting them and creating the required events when a user or a role changes.

### DIFF
--- a/src/Skoruba.IdentityServer4.Admin.EntityFramework.Identity/Repositories/IdentityRepository.cs
+++ b/src/Skoruba.IdentityServer4.Admin.EntityFramework.Identity/Repositories/IdentityRepository.cs
@@ -284,18 +284,20 @@ namespace Skoruba.IdentityServer4.Admin.EntityFramework.Identity.Repositories
 
         public virtual async Task<int> DeleteUserClaimsAsync(string userId, int claimId)
         {
+            var user = await UserManager.FindByIdAsync(userId).ConfigureAwait(false);;
             var userClaim = await DbContext.Set<TUserClaim>().Where(x => x.Id == claimId).SingleOrDefaultAsync();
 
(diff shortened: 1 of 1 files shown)


maulik-modi:
@skoruba , I am also looking for this functionality to capture events. Can you approve the merge?

skoruba:
Sure, please take a look at my comments - is it necessary to use `ConfigureAwait(false)`. cc @TomasHubelbauer

maulik-modi:
I think answer is it depends - 
>  If you have code in a library that may also run in a UI app, or legacy ASP.NET app, or anywhere else there may be a context, then you should still use ConfigureAwait(false) in that library.


https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html

skoruba:
OK, but is it required in this case? :)

maulik-modi:
I do not think Identity server Admin libraries being used outside Web API or Server side MVC. Hence, ConfigureAwait is unnecessary.

skoruba:
I have same opinion. Can you please @gingters remove usage of `ConfigureAwait(false)`? - or I can do it manually. 😊
Thanks guys!

PR #305, opened 2019-07-02
Implement multitenant support

chadbengen:
Added support for a multitenant configuration.  This includes moving some of the startuphelper code into builder extensions for easier configuration; however, the builder extensions need to be cleaned up and organized.  They work, but they do not feel clean.

Each web project's startup has commented code for configuring a AddSingleTenantConfiguration  (default), and AddMultiTenantConfiguration. 

I want to get this out there in case there's pieces i failed to consider or if there are some better/alternative ways to implement some of code.

**To configure for multi tenant:** 

1. Set each startup to AddMultiTenantConfiguration.
2. Set the layout page in STS to use SignInManager< MultiTenanUserIdentity >.
3. The Admin Program file should have seed = true, and seedMultiTenant = true.
4. Create the migration file by running the following from package manager console: 

Add-Migration AspNetMultiTenantIdentityDbInit -context MultiTenantUserIdentityDbContext -output Data/Migrations/Identity

The seed tenant uses company code of "0000" - four zeros.

**Points of interest:**

- MultiTenantUserDbContext was added along with additional dtos and interfaces.  These items are injected using the builders. In order to properly configure the dbcontext I copied the OnModelCreating from the aspnet identity repository and removed unique index constraints.
- In order to allow multiple emails/usernames and to require a tenant code for sign in there are new SigninManager, UserManager, etc. located with the new dbcontext in EntityFramework.Shared.  These are just wrappers.
- New user validators are also defined and the builder allows for adding addtional validators if required.  
- These new Identity objects are only injected if using Multi Tenant.
- The login controller and view are using some conditional logic to determine flow. 
- In order to get the tenant information as a claim there is an IProfileService object which sets tenantname and dbname as claims.  THese can be changed as necessary.  For our implementation these claims are required on the client and then used to construct the dbcontext.  Our implementation for multi tenancy is to have one db per tenant, except users.  If it's desired to have one db for all tenants then the IProfileService may need to be modified to include the tenantId as a claim so it can be used by the client dbcontext.


skoruba:
Hi @chadbengen 
Thank you for your effort for this PR. I take a time for review these changes and test them as well. 👍

src/Skoruba.IdentityServer4.STS.Identity/Controllers/AccountController.cs
```diff
@@ -129,10 +131,30 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
 
             if (ModelState.IsValid)
             {
-                var user = await _userResolver.GetUserAsync(model.Username);
+                TUser user;
+                //TODO: Code smell.  Evaulate if this huge method can be refactored.
+                if (_isMultiTenant)
+                {
+                    user = await _userResolver.GetUserAsync(model.TenantCode, model.Username);
```
pandarun: probably bug here, right signature is 
`
public async Task<TUser> GetUserAsync(string login, string tenant = null)
`
so should be
`
    user = await _userResolver.GetUserAsync(model.Username, model.TenantCode); 
`

src/Skoruba.IdentityServer4.STS.Identity/Helpers/UserResolver.cs
```diff
@@ -15,17 +16,36 @@ public UserResolver(UserManager<TUser> userManager, LoginConfiguration configura
...
-                case LoginResolutionPolicy.Username:
-                    return await _userManager.FindByNameAsync(login);
-                case LoginResolutionPolicy.Email:
-                    return await _userManager.FindByEmailAsync(login);
-                default:
-                    return null;
+                switch (_policy)
+                {
+                    case LoginResolutionPolicy.Username:
+                        return await _userManager.FindByNameAsync(login, tenant);
```
pandarun: probubly bug here, signature is
`
public Task<TUser> FindByEmailAsync(string tenantCode, string email)
 `
so should be
`
return await _userManager.FindByNameAsync(tenant, login);
`

src/Skoruba.IdentityServer4.STS.Identity/Helpers/UserResolver.cs
```diff
@@ -15,17 +16,36 @@ public UserResolver(UserManager<TUser> userManager, LoginConfiguration configura
...
+
+                    case LoginResolutionPolicy.Email:
+                        return await _userManager.FindByEmailAsync(login, tenant);
```
pandarun: and here, args are swapped.

src/Skoruba.IdentityServer4.Admin.EntityFramework.Shared/Managers/MultiTenantUserManager.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        [Obsolete]
+        public override Task<TUser> FindByEmailAsync(string email)
+        {
+            return null;
+        }
+
+        public Task<TUser> FindByEmailAsync(string tenantCode, string email)
+        {
+            return (Store as IMultiTenantUserStore<TUser>).FindByEmailAsync(tenantCode, email);
```
pandarun: possible NRE

src/Skoruba.IdentityServer4.Admin.EntityFramework.Shared/Managers/MultiTenantUserManager.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public Task<TUser> FindByNameAsync(string tenantCode, string userName)
+        {
+            return (Store as IMultiTenantUserStore<TUser>).FindByNameAsync(tenantCode, userName);
```
pandarun: IMultiTenantUserStore receives normalizedKey, so somewhere should be call to _userManager.NormalizeKey

src/Skoruba.IdentityServer4.Admin/DependencyInjection/BuilderExtensions.MultiTenant.cs
```diff
@@ -0,0 +1,78 @@
...
+            (this IServiceCollection services,
+            IHostingEnvironment hostingEnvironment,
+            Action<DbContextOptionsBuilder> identityDbContextOptions,
+            Action<ConfigurationStoreOptions> configurationStoreOptions,
+            Action<OperationalStoreOptions> operationalStoreOptions,
+            Action<DbContextOptionsBuilder> logDbContextOptions,
+            Action<IdentityOptions> identityOptions)
+        {
+            var builder = new Builder(services);
+            builder.Services.AddMemoryCache();
```
pandarun: probably should be distributed cache, otherwise how to guarantee coherency between this cache and src/Skoruba.IdentityServer4.STS.Identity/DependencyInjection/BuilderExtensions.MultiTenant.cs MemoryCache

src/Skoruba.IdentityServer4.STS.Identity/Controllers/AccountController.cs
```diff
@@ -129,10 +131,30 @@ public async Task<IActionResult> Login(LoginInputModel model, string button)
 
             if (ModelState.IsValid)
             {
-                var user = await _userResolver.GetUserAsync(model.Username);
+                TUser user;
+                //TODO: Code smell.  Evaulate if this huge method can be refactored.
+                if (_isMultiTenant)
```
pandarun: maybe it would be better to have one path? e.g all apps would be multi-tenant and have tenant by default or create two separate project templates (single and multiple tenants), otherwise, only one branch will be executed.

src/Skoruba.IdentityServer4.STS.Identity/DependencyInjection/BuilderExtensions.MultiTenant.cs
```diff
@@ -0,0 +1,109 @@
...
+                    (identityOptions)
+                .AddMultiTenantIdentityServer
+                    <MultiTenantUserIdentity, IdentityServerConfigurationDbContext, IdentityServerPersistedGrantDbContext>
+                    (configuration, hostingEnvironment, logger, identityServerOptions, configurationStoreOptions, operationalStoreOptions)
+                .AddMvcWithLocalization
+                    <MultiTenantUserIdentity, string>();
+
+            builder.Services.Remove(builder.Services.FirstOrDefault(d => d.ServiceType == typeof(IUserValidator<MultiTenantUserIdentity>)));
+            builder.AddUserValidator<MultiTenantUserIdentity, MultiTenantUserValidator>();
+            builder.AddUserValidator<MultiTenantUserIdentity, RequireTenant>();
```
pandarun: How one could register, should it choose tenant code beforehand like in login form or tenant should be automatically provided per new user?

ganmuru: Tenant should be decided beforehand, depending on for which client the user is trying to register.
Usually the subdomain of the client url should provide the info for selecting Tenant. It can also be decided using Request headers.
Example:
Client: xyz.clientA.com and in IS4 client redirect URI: xz.clientA.com/signin-oidc
then process flow would be 
xyz.clientA.com => IS4 - Register UI on register get the tenant code based on subdomain "xyz".

skoruba:
#470 - closed because of this new PR.
Thanks

PR #317, opened 2019-07-16
Feature/windows support

aiscrim:
Added all the logic needed to challenge the Windows authentication. Inspiration taken from IdentityServer4.QuickStart.UI project, but heavily refactored to better integrate with this project.

Added some settings in the STS to configure the support to the Windows authentication:

- **_AutomaticWindowsLogin_**: if true, when authentication is required and if the request comes from the same network as the STS, no login screen is shown and the Windows login automatically takes place;
- **_IncludeWindowsGroups_** (moved from AccountOptions): if user uses Windows authentication, should we load the groups from Windows?
- **_WindowsGroupsPrefix_**: if we load the groups from Windows, load only the groups that start with this prefix
- **_WindowsGroupsOURoot_**: if we load the groups from Windows, load only the groups under this OU.

Added user photo to the profile: when the user logins through Windows it is read from the Active Directory, otherwise the "picture" claim of the external login is checked.
When the user photo needs to be shown, we first check if the "picture" claim exists, otherwise we use the Gravatar picture as before.


skoruba:
I will take a look ASAP, currently I am on vacation. 😊
Thanks for your work on this! 👍🏼

aiscrim:
Don't worry, enjoy your holidays! I plan to open some more PRs before I start mine, so stay tuned! :-)

skoruba:
Do you have some experience how to solve the use case if user comes from intranet network then use windows auth otherwise show some classic form which will be connected to AD via LDAP?

aiscrim:
@skoruba  Sorry for the delay, it was my turn for the holidays! :)
I don't have experience on the exact scenario you describe, but I can give a look. Just to be sure we are on the same page, when you say "classic form" you mean something like the Local Login form or the browser dialog popping up when the Windows authentication is required but not provided?

skoruba:
I mean local login. 😊

aiscrim:
Ok, that's what I thought. I have done some ugly tests locally and I managed to have a working prototype!
In the coming days I will refactor the code and update this PR with a new commit so that the user will be presented a login form if they ask for Windows authentication and they are not in the same network as the Identity Server.

skoruba:
Nice, thanks!

aiscrim:
Ok, I did it! Please give a look, in the tests I did it seems to work smoothly. 
I added localization for the new form, but only in English due to my ignorance of the other supported languages.

navidminapour:
Hello,

Why don't these changes exit in the latest version?
For example, LoginConfiguration class just have one property.

aiscrim:
If you refer to the main branch, Jan needs to review the code and merge this feature.
Btw please note that the latest commit of this feature branch moves the configuration of the Windows authentication to a separate class, WindowsConfiguration, so in LoginConfiguration you will still see only one property. 😊

skoruba:
Thanks for this awesome PR, I will continue with this feature in branch - `feature/windows-auth`.

PR #319, opened 2019-07-17
Configurable localization

aiscrim:
Not all the available cultures are needed or desirable in every organization. With these changes:

- It is now possible to configure in the appsettings the subset of cultures that we want to use in our instance. 
- The default culture can be configured too.
- Adding a new available culture is cleaner: now it is just a matter of adding a new element to the CultureConfiguration.AvailableCultures array, instead of having to change the code in the StartupHelper class.
- When there is only one supported culture, no language selection combo is shown at all.

These changes have been made both in the STS and in the Admin applications.


skoruba:
I've merged another PR with a lot of changes in configuration - can you please check these conflicts? 👍 I can do it manually if you want. 😊
Thanks for your great work!

aiscrim:
Done! Please merge this PR to dev before any other PR that changes the configuration, to avoid doing the same activity again and again. :)

PR #178, opened 2019-02-19
Add Docker support

Seaear:
1. Add "127.0.0.1 id.example.com" to hosts file.
2. In the directory where docker-compose.yml, Use the command "docker-compose build" build images, and use "docker-compose up" start services.

PR #325, opened 2019-07-30
Feature/docker support seaear base

rherlt:
- Used the very nicely created code from @Seaear from the branch [feature/docker-support](https://github.com/Seaear/IdentityServer4.Admin/tree/feature/docker-support) and resolved the merge conflicts from #178 
- Added new app setting "IdentityRequireHttpsMetadata" (in Admin and Admin.Api) which optionally disables the check for HTTPS metadata, to be able to run this in a kubernetes cluser with enabled [TLS-termination](https://en.wikipedia.org/wiki/TLS_termination_proxy) (e.g.  in the load balancer or in an api gateway)
- Updated seaear's docker-compose file to use 127.0.0.1.xip.io instead of _id.example.com_ mapped to 127.0.0.1 in the host file


skoruba:
Thank you, I will try it. 👍🏼

rherlt:
- yesterday I added a handling for , see [forward the scheme for linux and non IIS reverse proxies](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.2#forward-the-scheme-for-linux-and-non-iis-reverse-proxies)  for details. Without that, the urls in the discovery document had the wrong url scheme (http instead of https).

brinehart:
@rherlt Why is the Admin.Api Dockerfile necessary? I might be missing where it is being called in the docker-compose?

rherlt:
> @rherlt Why is the Admin.Api Dockerfile necessary? I might be missing where it is being called in the docker-compose?

Hi @brinehart,

the docker file in the _Admin.Api_ is necessary to build a docker image for that application. You are right, this is currently not included in the docker-compose file.

The main reason ist, that I did not finish the implementation for mayself so far. Until now, I was just concentrating on the _STS.Identity_ and _ the _Admin.Api_. Both Docker imges are already run in a Kubernetes cluster.

Although the build of the _Admin.Api_ Docker image works, I did not find the time to test this so far. I hope I will find the time to finish the Kubernetes of setup for the _Admin.Api_ during the next week. Then I could easily add this also to the docker-compose file.

Regards,
Rico

skoruba:
Did you have a time for including Admin.Api into docker-compose? :)

skoruba:
@rherlt - Thanks for your great PR, we will continue in PR - #367, which contains also support for Api. Feel free to send some improvements in the future.

PR #333, opened 2019-08-07
Support for single quotes in PickerTag labels.

nlz242:
Currently, if an apostrophe is used in a label of the PickerTag, it generates invalid html. 
It is invalid because the generated div has a data-bind attribute which has it's value inside single-quotes. This is obviously done because the value contains JSON, thus it has many double-quotes and should not be encased in double-quotes. 
 
Fix escapes characters, like the single quote.
Useful for french, which has apostrophe in words.
For example : "The item" becomes "L'item"


nlz242:
Had not seen that everyone is doing their pull request against the dev branch. 
Going to re-do the pull request against the dev branch.

Issue #343, opened 2019-08-14
DatePicker's locale doesn't change when switching language


nlz242:
In the admin UI, in the logs page, the DatePicker's language is always english (ignores the selected language). 

For example, if i select Russian, the calendar still shows "August" and all the abreviated days in english: 

![image](https://user-images.githubusercontent.com/3332745/63026645-8ef12780-be79-11e9-98a6-c58c8ba8ad4b.png)

The calendar comes with Locales for most cultures and they can be found in Skoruba.IdentityServer4.Admin\node_modules\bootstrap-datepicker\dist\locales 
Probably just have to include them in the gulp script and set the language on the picker as per the documentation here:  https://bootstrap-datepicker.readthedocs.io/en/latest/i18n.html

skoruba:
I am closing this issue, it was released in master branch.

PR #344, opened 2019-08-14
Add language support to DatePicker

nlz242:
Fix for #343 
Adds localisation to the Log's DatePicker.
Tested in all currently supported languages + incomming french.
Defaults to english if no AspNetCore cookie found.

Support for chinese is a bit janky as the picker supports two different chineses variants but doesn't have a neutral one (no "zh", only "zh-CN" and "zh-TW"). 

GetCookie function could be replaced by a jQuery plugin but i did not want to alter this project too much, as i am a new contributor.


nlz242:
Woops. Did that against the master branch, my bad. Going to redo against dev.

PR #346, opened 2019-08-15
Initial upgrade for .NET Core 3.0 and CORS

sabitertan:
This PR includes minimum upgrade path explained by [Migrate from ASP.NET Core 2.2 to 3.0](https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio).
It also includes nuget package upgrades to the latest previews.


skoruba:
Nice, this is cool 👍😊 I look forward to .NET Core 3.0.
Thanks!

sabitertan:
I updated Fluent Assertions in this PR but seems like tests needs to be updated :/

skoruba:
Yes, I know, I have older version of FluentValidation in UnitTests because I will need to rewrite probable a lot of tests, so upgrading of this library will be later. :)

skoruba:
I will continue here: #390.
Thanks for initial version!

Issue #351, opened 2019-08-22
Error management when editing profile


aiscrim:
When something wrong happens in the ManageController/Index action (e.g. the user changes the email address but it's already in use by another user) an exception is thrown, resulting in a 500 response. Moreover, the details of the error are not clear enough, even locally: in the 'email address already in use' example above, all I get is:

> ApplicationException: Unexpected error occurred setting email for user with ID ...

And nothing goes to the logs, so there's no way of understanding what's the problem apart from debugging it in VS.

For me it's quite urgent to solve this issue, so if you are busy I can fix it myself and open a PR, just let me know.

skoruba:
Please, open a PR if you can, because I am currently busy.
Do you try to setup email which is used for another user? I'll be happy for any fix/details on this. 😊
Thanks!

skoruba:
I think this line is your issue: https://github.com/skoruba/IdentityServer4.Admin/blob/master/src/Skoruba.IdentityServer4.STS.Identity/Controllers/ManageController.cs#L80
Please, check the result in this line.
Thanks.

aiscrim:
Yes, exactly, and as you see when that call fails an exception is thrown. Instead I think it will be better to return a user-friendly message. I'll do it!

skoruba:
You are right, in this case will be better to have some nice message. :)
Thanks @aiscrim

aiscrim:
Working on the fix I discovered a few other related flaws in the ManageController:

- [Here](https://github.com/skoruba/IdentityServer4.Admin/blob/master/src/Skoruba.IdentityServer4.STS.Identity/Views/Manage/_StatusMessage.cshtml#L5), the way StatusMessage is classified as success or danger is wrong: not all the errors start with the word "Error" (actually, most of the errors start with "Unexpected error..." and they would be classified as success), and anyway in languages other than English that would never work. I suggest replacing the string StatusMessage with an object (bool IsSuccess + string Message).
- TwoFactorAuthenticationViewModel is missing the StatusMessage field, so even if the message is set in the calling action, it never gets displayed.
- Some messages contain the user id, so they seem to be intended more for the log than for the UI, but they are displayed as StatusMessage. The pattern used for the PasswordChanged/PasswordChangedLog messages should be used everywhere imho.

If you agree I can fix the first two issues in the same PR. The last one is a bit trickier for me to solve because it involves creating new localized strings in all the supported languages, so I'd just ignore it atm.

PR #354, opened 2019-08-27
Improvements in error handling

aiscrim:
A few changes in the way the errors are handled, mainly in the profile management. This PR tackles the issue #351 

- If some error occurs updating the user profile, the error is now shown to the user
- In the ManageController, the status message is now a class and it's possible to distinguish between error or success messages using the IsSuccess property. Before, a message was considered an error only if it started with the word "Error"
- Added Status property to TwoFactorAuthenticationViewModel, so any feedback to the user can be displayed in the 2fa page too
- SendVerificationEmailBtn is hidden if the user email address was changed but not saved yet, to avoid confusion
- Updated UserIsAbleToUpdateProfile integration test because now because of the changes above when the profile is correctly updated the result is not a redirect but a 200 success code
- If an error occurs sending an email, the application notifies the user instead of just crashing

Some messages were added in the English localization, it would be great to have them translated in the other languages too.


skoruba:
Thank you for your PR, let me review it. 😊

skoruba:
I merge this great PR into branch `feature/improvements-in-error-handling` - I will continue in this branch with this feature, adding of missing translation etc. 😊
Thanks again for your great PRs and I apologize for my delay with processing them.

PR #361, opened 2019-09-18
 change primary key in asp.net identity

mehradi-github:
change primary key in asp.net identity to GenericType


skoruba:
Hi, why do you change primary key to long?
Thanks

mehradi-github:
Hi skoruba Dear,
Defulat primary key was string and didn't change to long or ohter type

mehradi-github:
I changed, Because ASP.NET Identity had used **GenericType**

skoruba:
Yes, but default value is string which is automatically generated Guid on background and this value is recommended. I had by default INT value and I changed this value back to string - GUID. :)

mehradi-github:
Yes ,But  I should use to **TKey** in code, because is filxeable
and it change to GUID,long,string and etc easily

mehradi-github:
Searching Strings in SQL Server is Expensive

skoruba:
It is a good idea with `TKey`, but I need split your PR into smaller parts, because it is complex PR. I will probable merge it manually and use same logic in MVC project as well. But I will definitely use the guid - like primary key of Identity model - which it's default.
Thanks for your work! 👍

dongfo:
Trust me, no one wants to use string as a primary key. Because the performance is really bad

PR #372, opened 2019-10-07
Feature/deny self delete

aiscrim:
Users should not be able to delete their own account: if administrators are allowed to delete their own account, they could be left locked out.
This way we ensure that at least one administrator is always present in the system.


b0:
Hi,

this PR may have introduced an unexpected behavior...

Currently, API can be called by client on behalf of a user, and by standalone client. 
This is so because in Authorization policy we are looking for **Role** or **Client_Role** claim and I find this **correct**. 
In my case, I have a machine accessing API - there is no User clicking anything(I am importing, deletging users), and in another case I have a User clicking and accessing API. Both valid cases...

Problem is that code checks for User's SubjectId which is not present when API is accessed as a machine client... 

@aiscrim Do you have some thoughts about this?

Thank you,
b0

aiscrim:
Hi @b0,
I think you're right, I overlooked the machine-to-machine scenario here. I'll open a new PR to patch this one. If it's urgent for you, please feel free to open it yourself: I wouldn't be able to work on it before next week probably.

skoruba:
Hi @b0 - thanks for reporting this issue.

PR #380, opened 2019-10-21
Batch import of all AD users

aiscrim:
There are cases when we need to have all the Windows users registered in the database, so that for example calling the Admin API to retrieve all the users we get all the AD users instead of just those who logged in at least once.
I have added a button in the Identity Server top menu to launch a massive import from AD, but I am not 100% sure this is the best place: probably putting it in the Admin menu or even calling it during the seed operation would be a better option, but for the moment it was easier for me to leave it in the IdentityServer project (all the Windows auth classes are there).
I open this PR looking for some feedback and ideas to refactor it.


skoruba:
I am merging this into windows-auth branch and I will continue there with next improvements, but again, thanks for your great work! :-)

Issue #108, opened 2018-11-05
Host IdentityServer4.Admin on Azure


FrankDoersam:
Hello,

the project looks really great. How should I do that the service is hosted in Azure?

Yours sincerely
Frank

skoruba:
Hi Frank,
I will prepare the guide - how to create the “production” release. :)
I will contact you.

Thanks!

FrankDoersam:
Hello skoruba,

thank you for the fast feedback. I would really appreciate it if you could show how to fully publish the IdentityService in Azure. I wanted to use this for an Alexa skill :).

Yours sincerely
Frank

skoruba:
Please take a look at: https://github.com/skoruba/IdentityServer4.Admin/blob/master/docs/Configure-Azure-Deploy.md
Thanks

tsangian:
Hi Skoruba,

I have no problem running on local and then I tried to deploy to Azure App Service using Visual Studio Publish feature. Before deploying I followed the [documentation](https://github.com/skoruba/IdentityServer4.Admin/blob/master/docs/Configure-Azure-Deploy.md), but then it shows "Internal Server Error: An error occurred while starting the application." when I tried to access the STS.Identity UI. Here's the exception from Log table in database (Azure SQL):

System.InvalidOperationException: IDX20803: Unable to obtain configuration from: '[PII is hidden]'. ---> System.IO.IOException: IDX20804: Unable to retrieve document from: '[PII is hidden]'. ---> System.Net.Http.HttpRequestException: Response status code does not indicate success: 500 (Internal Server Error).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at Microsoft.IdentityModel.Protocols.HttpDocumentRetriever.GetDocumentAsync(String address, CancellationToken cancel)
   --- End of inner exception stack trace ---
   at Microsoft.IdentityModel.Protocols.HttpDocumentRetriever.GetDocumentAsync(String address, CancellationToken cancel)
   at Microsoft.IdentityModel.Protocols.OpenIdConnect.OpenIdConnectConfigurationRetriever.GetAsync(String address, IDocumentRetriever retriever, CancellationToken cancel)
   at Microsoft.IdentityModel.Protocols.ConfigurationManager`1.GetConfigurationAsync(CancellationToken cancel)
   --- End of inner exception stack trace ---
[...]

PR #386, opened 2019-10-31
Update Configure-Azure-Deploy.md

killerrin:
I ran into this issue when deploying to Azure. The solution is written inside of #108 #issuecomment-538904975 https://github.com/skoruba/IdentityServer4.Admin/issues/108#issuecomment-538904975

Adding to documentation to solve an hour of frusturation for people following the tutorial


skoruba:
Thanks! :-)

PR #394, opened 2019-11-08
Added Danish translation into Localization

HrDahl:
Added support for Danish language in the solution. 

Thanks for the great project :)


skoruba:
Thank you for your great job 👍🏼

PR #367, opened 2019-09-28
docker support

ChrisSzabo:
- Added support for building docker containers for Admin, Api, and STS projects. Based off of code from @rherlt (https://github.com/skoruba/IdentityServer4.Admin/pull/325). 

- This version is a little more efficient when building the images because the newer VS docker tooling generated dockerfile code that also copies project references (.csproj files) when creating the dependency cache layers.

To run:
docker-compose up --build

To stop:
docker-compose down

Api Url: http://127.0.0.1.xip.io:5000/swagger/index.html
IdentityServer Admin: http://127.0.0.1.xip.io:9000
IdentityServer : http://127.0.0.1.xip.io

PR #395, opened 2019-11-10
docker support with .NET Core 3.0

### LINKED ISSUE #121: Docker support
Hey guys!
I tried to pack the project into docker to make it easy to deploy and faced some problems, maybe some one could help here. Changes published to [docker branch](https://github.com/NTurakulov/IdentityServer4.Admin/tree/docker)

**Requirement**
Make it easy to deploy Identity Server Admin to use it in different apps as a part of micro service architecture. I saw somewhere in the feed that @skoruba planned to pack the project into NuGet but I think docker another way to make it portable

**Solution**
Pack the project (Admin and STS) into docker images and run them at any time with required port inside your cluster

**Implementation**
I added two Dockerfiles in target project folders, docker-compose.yml in root plus several powershell scripts in the folder with same name. It works but made dummy. To start run containers - run 'run-docker.ps1' script or execute commands from it one by one (for non-Windows platforms).

SQL server is also packed into an image and run as a part of compose with persistent external volume (skoruba-identityserver-sql). All three containers (SQL, STS, Admin) work within same docker network (skoruba-identityserver-network) and can interact with each other

**Note** 
- Admin project port 80 is mapped to host's port **5001** instead of 9000
- MS SQL container's port 1433 is mapped to host's 1433**3**! This made for compatibility with MS SQL instance on the host

**Problems**

- Dependency projects are not packed as NuGet and cannot be restored with 'dotnet restore'. Because of that instead of using single multi step Dockerfile I build it on host then just copy to aspnetcore-runtime image. This can be solved with modified Dockerfile which copies 'bin' folder with dependencies but prebuild on host is required anyway. Also projects use 'npm install' which is not preinstalled into default aspnetcore-sdk build image.
    - As dotnet CLI calls required before 'docker build' they're packed into build-docker-*.ps1 scripts
- I've added DockerRelease config file with overriden parameters but they don't apply so duplicated in docker-compose.yml and passed with 'environment'
- While STS and Admin containers can use short docker-compose.yml service names when reference to sql container successfully I can't access to sts/.well-known/openid-configuration from within Admin project
- Don't know how to apply EF migrations from within container. Before building docker image DbContext reference to host DB, after running a container not sure how to do that. Maybe something like 'EXEC dotnet ef database update'? Need to try this way
- Both projects are configured to run in 'Development' environment because certificate for HTTPS is not set up

bravecobra:
I've made this PR based on #367 by @ChrisSzabo. The amount of work to get it all working properly seemed to be more extensive given the branch was getting rather old.  I merged in the latest dev branch, containing the latest .NET Core 3.0 support (#390 ) and the Audit logging (#391 )

So what's changed?

- Fixed the IdentityServerBaseUrl for Swagger API as mentioned in #367 
- Fixed Swagger redirect URI misconfiguration
- Fixed the outstanding merge conflicts of #367 
- Merged the latest dev branch
- Changed the environment variable to use double underscore. It worked in VS2017 with colons, but VS2019 seems to require it in order to debug properly. [docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.0#keys) and [SO](https://stackoverflow.com/questions/55533445/environment-variables-null-in-new-visual-studio-2019-but-not-visual-studio-2017)
- Added a `docker-compose.vs.debug.yml` and `docker-compose.vs.debug.yml` to enable debugging with a seeded environment.

We'll need to update the docs, but I suggest getting this in first.

So now we have the following possibility to get a running seeded and debug-able (in VS) environment:

```powershell
docker-compose build
docker-compose up -d
```

This PR closes #121


skoruba:
Amazing work @bravecobra - thanks a lot 😊 I will test it ASAP 👍

bravecobra:
Don't forget that this still assumes you have the proper migrations generated in place. We can fix that later in the build pipeline I've prepared with github actions. One step at the time...

xmichaelx:
Looks solid - i've checked the changes in file. Later today I'll download, build and test it but I think that we want to merge it as soon as possible.

xmichaelx:
Looks ok, let's merge it @skoruba

Issue #408, opened 2019-11-16
Adding support for multiple DbProviders


bravecobra:
I noticed lots of issues being reports regarded missing database migrations. 

In EF Core it is possible add migrations for multiple DbProviders: <https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/providers?tabs=dotnet-core-cli>

So instead/on top of having the end-user having to generate the migrations for him/herself, we can provide those migrations already and allow for choosing the correct DbProvider through configuration. This would also allow for upgrading existing docker containers. 
We should also allow the end-user to provide his/her own set of migrations, should they choose to do so. There might a couple of rogue migrations in existence with custom changes. All that through configuration of course.

skoruba:
I am glad that you suggest this idea, because I am working on this locally currently. I will add the ef migrations into source control for multiple db providers, because it will help for a lot of use cases.
It will be possible to switch among db providers from appsettings.json.

maxrev17:
As a newbie to this repo I've been working through getting this to run in docker and have stumbled with the migrations as my machine is bare and I'm relying on docker for .net etc. This would be an awesome change to speed up getting it up and running out of the box :).

skoruba:
I am closing this issue, it was released in master branch.

PR #414, opened 2019-11-20
Support for multiple db providers

skoruba:
I added the following projects into solution:

- Skoruba.IdentityServer4.Admin.EntityFramework.MySql
- Skoruba.IdentityServer4.Admin.EntityFramework.PostgreSQL
- Skoruba.IdentityServer4.Admin.EntityFramework.SqlServer

These new projects also contain initial migrations, so I need update readme file, because currently it is not necessary to generate migrations manually, it will be a part of this project as well.

In `appsettings.json` is possible to switch between db providers:
```
"DatabaseProviderConfiguration": {
        "ProviderType": "SqlServer"
    }
```

`ProviderType` has keys:
- MySql
- PostgreSQL
- SqlServer

Default db provider is defined `SqlServer`.

Please, check this out: @TomasHubelbauer @xmichaelx @aiscrim @bravecobra @maxrev17
Solved this task: #408


build/add-migrations.ps1
```diff
@@ -0,0 +1,36 @@
...
+
+        dotnet ef migrations add $migration -c AdminIdentityDbContext -o Migrations/Identity -p $dbProviders[$key]
+        dotnet ef migrations add $migration -c AdminLogDbContext -o Migrations/Logging -p $dbProviders[$key]
+        dotnet ef migrations add $migration -c IdentityServerConfigurationDbContext -o Migrations/IdentityServerConfiguration -p $dbProviders[$key]
+        dotnet ef migrations add $migration -c IdentityServerPersistedGrantDbContext -o Migrations/IdentityServerGrants -p $dbProviders[$key]
+        dotnet ef migrations add $migration -c AdminAuditLogDbContext -o Migrations/AuditLogging -p $dbProviders[$key]
+    }
+}
+
+Remove-Item appsettings.json
```
TomasHubelbauer: Why is this backup thing there? Does generating the migrations write to appsettings or is there another source of risk of losing the data in this file? If yes, could you add this to a comment?

src/Skoruba.IdentityServer4.Admin.Api/Helpers/StartupHelpers.cs
```diff
@@ -124,43 +126,29 @@ public static void AddLogging(this IApplicationBuilder app, IConfiguration confi
...
+                    break;
+                case DatabaseProviderType.PostgreSQL:
+                    services.RegisterNpgSqlDbContexts<TIdentityDbContext, TConfigurationDbContext, TPersistedGrantDbContext, TLogDbContext, TAuditLoggingDbContext>(identityConnectionString, configurationConnectionString, persistedGrantsConnectionString, errorLoggingConnectionString, auditLoggingConnectionString);
+                    break;
+                case DatabaseProviderType.MySql:
+                    services.RegisterMySqlDbContexts<TIdentityDbContext, TConfigurationDbContext, TPersistedGrantDbContext, TLogDbContext, TAuditLoggingDbContext>(identityConnectionString, configurationConnectionString, persistedGrantsConnectionString, errorLoggingConnectionString, auditLoggingConnectionString);
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(databaseProviderType));
```
TomasHubelbauer: Would be cool to include the valid values in this exception.

src/Skoruba.IdentityServer4.Admin.EntityFramework.Shared/Configuration/DatabaseProviderConfiguration.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Skoruba.IdentityServer4.Admin.EntityFramework.Shared.Configuration
+{
+    public class DatabaseProviderConfiguration
+    {
+        public string ProviderType { get; set; }
```
TomasHubelbauer: Is it valid to use the enum type here or does this have to be a string in this scenario?

build/add-migrations.ps1
skoruba: Original appsettings.json is stored is appsettings-backup.json - because is modified in script and finally is removed and restored from backup file. :-)

src/Skoruba.IdentityServer4.Admin.EntityFramework.Shared/Configuration/DatabaseProviderConfiguration.cs
skoruba: Done, good idea 👍

src/Skoruba.IdentityServer4.Admin.Api/Helpers/StartupHelpers.cs
skoruba: Any idea how to do it?

TomasHubelbauer: This exception has a secondary constructor which accepts a custom message:

https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=netframework-4.8#System_ArgumentOutOfRangeException__ctor_System_String_System_String_

You could call something like:

```cs
throw new ArgumentOutOfRangeException(nameof(databaseProviderType), $"The value needs to be one of {string.Join(", ", Enum.GetValues(typeof(ProviderType)))}.");
```

skoruba: Great idea, I used your code, but I changed Enum.GetValues to Enum.GetNames.

PR #418, opened 2019-11-25
Update RedirectUris

Varorbc:


skoruba:
Why do you change the uri? I think original uri is correct. Thanks

Varorbc:
http 5000
https 5001

skoruba:
Check what you changed in PR.

Varorbc:
https://github.com/skoruba/IdentityServer4.Admin/blob/5cda5fd65095eb99563e82360bbdaaad49433b3a/src/Skoruba.IdentityServer4.STS.Identity/Properties/launchSettings.json#L24

PR #420, opened 2019-11-26
Initial IStringLocalizer based localization of ASP.NET Core Identity

duki994:
@skoruba 

A simple implementation of ASP.NET Core Identity translation via inbuilt IStringLocalizer.
I have already used these in several projects.

I can add another layer of abstraction (for example IIdentityLocalizer/IIdentityErrorMessageLocalizer) so consumers can implement it independently, and pull that out into  
_(namespacebefore).AspNetCore.Identity.Localization_

Opens up multiple implementations not coupled to ASP.NET Core localization implementation.
It may be reinventing wheel, though it's quite easy thing to do. 

What's your opinion about this? :)


duki994:
@skoruba 

I just  added unit tests to cover default _IdentityErrorMessages_ localization implementation.

duki994:
@skoruba @TomasHubelbauer @xmichaelx 
Closing this to open same pull request from correct branch. I made myself issue when commiting in dev directly on my fork

PR #422, opened 2019-11-26
Refactor STS

fatalistt:
1. Fix typos
2. Remove extra configuration interfaces and service provider
3. Throw exceptions rather logging errors while setting Signing Credentials, cause without that we will get a non-working app (it'll throw an exception that there is no signing credentials)
4. Throw exceptions rather logging errors while setting Validation Key, cause we get a bad configured app if we will continue.
5. Refactor startup and program.cs to be more ".net core 3"-way.

### DIFF
--- a/src/Skoruba.IdentityServer4.STS.Identity/Program.cs
+++ b/src/Skoruba.IdentityServer4.STS.Identity/Program.cs
@@ -23,7 +23,6 @@ namespace Skoruba.IdentityServer4.STS.Identity
                 {
                     webBuilder.ConfigureKestrel(options => options.AddServerHeader = false);
                     webBuilder.UseStartup<Startup>();
-                    webBuilder.UseSerilog();
                 });
     }
 }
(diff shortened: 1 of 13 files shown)


src/Skoruba.IdentityServer4.STS.Identity/Program.cs
```diff
@@ -8,14 +9,21 @@ public class Program
...
+                .ConfigureLogging((hostingContext, logging) => {
+                    logging.ClearProviders();
+                    var logger = new LoggerConfiguration().ReadFrom.Configuration(hostingContext.Configuration).CreateLogger();
+                    logging.AddSerilog(logger);
+                })
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.ConfigureKestrel(options => options.AddServerHeader = false);
+                    webBuilder.UseStartup<Startup>();
+                    webBuilder.UseSerilog();
```
duki994: @fatalistt 

You don't need _UseSerilog_ on _IWebHostBuilder_

fatalistt: Sorry for a long anwer, I was busy at work.
Yes, you're right, `IWebHostBuilder.UseSerilog` is redunant.

skoruba: Can you remove this line before final merge, please? :-)
Thanks

fatalistt: Already done that :)

skoruba:
Great work, thanks!

PR #423, opened 2019-11-26
IdentityErrorMessages - IStringLocalizer based default localization of ASP.NET Core Identity

duki994:
@skoruba @TomasHubelbauer 

A correct reopen of https://github.com/skoruba/IdentityServer4.Admin/pull/420

A simple implementation of ASP.NET Core Identity translation via inbuilt IStringLocalizer.
I have already used these in several projects.

I can add another layer of abstraction (for example _IIdentityLocalizer/IIdentityErrorMessageLocalizer_) so consumers can implement it independently, and pull that out into
_**(namespacebefore)**.AspNetCore.Identity.Localization_

Opens up multiple implementations not coupled to ASP.NET Core localization implementation.
It may be reinventing wheel, though it's quite easy thing to do.

What's your opinion about this? :)

### DIFF
--- a/src/Skoruba.IdentityServer4.Admin/Helpers/Identity/IdentityErrorMessages.cs
+++ b/src/Skoruba.IdentityServer4.Admin/Helpers/Identity/IdentityErrorMessages.cs
@@ -337,11 +337,11 @@ namespace Skoruba.IdentityServer4.Admin.Helpers.Identity
         }
 
         /// <summary>
-        /// Generates <see cref="IdentityError"/> with <see cref="IdentityError.Code"/> named as <paramref name="name"/>
+        /// Tries to generate <see cref="IdentityError"/> with <see cref="IdentityError.Code"/> named as <paramref name="name"/>
         /// And error description which uses <paramref name="name"/> as lookup key for <see cref="IStringLocalizer{T}"/>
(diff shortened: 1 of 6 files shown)


skoruba:
Thank you @duki994 for this PR, is it possible to setup some condition - if current translation will be missing - use default english?

duki994:
@skoruba 

Thanks for suggestion. I used two pronged approach (don't know if that's good phrase for this approach :P).

1. It uses IdentityErrorDescriber.resx (without culture name)

     This is based on official .NET Core  https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-3.0#culture-fallback-behavior

2. Case when approach one fails (localizer can't find resource)

    Code is defaulting to base methods in default IdentityErrorDescriber from ASP.NET Core Identity.


All cases are covered with unit tests.

skoruba:
This is perfect 👍🏼🙂

duki994:
@skoruba 
Thanks 💯 

I just added missing test data for PasswordRequiresUpper method.
That would be 100% public API coverage of IdentityErrorMessages with tests.

Merge when ready :)

skoruba:
Great work, thanks Dusan - @duki994. 👍

PR #434, opened 2019-12-07
Aspnetidentity localization fixes

duki994:
@skoruba 

My earlier pull request didn't take in account formatting of compound strings.
Now everything works.

I made test cases and made sure they assert that argument for error (email, role etc.) is contained within compound formatted string.


skoruba:
Thanks Dusan. 👍🏼

PR #436, opened 2019-12-10
WIP: Kubernetes setup on docker-for-desktop

bravecobra:
While still a work in progress, the current yml files will allow running the docker images on the single-node cluster on docker-for-desktop on Windows. It hasn't been tested on MacOS 
You will need to edit your hosts file adding `identityadmin.k8s.local` to point to the IP of the machine running the cluster. The shortcut we took in `docker-compose` will not work on k8s as DNS resolving needs to happen at a different level. 
I also added metallb to allow k8s to pass out external IP addresses all pointing at localhost. This should allow load balancing in future scenarios (by adding extra replicas).

This configuration is for dev usage only. Do NOT use it on production!! 
Missing features are: 
1. TLS support
2. Volume claims (to persist the data of sql server between reboots)
3. Configure the different services with config maps instead env variables
4. Store the database credentials in a k8s Secret. 

We need to package the docker images and store them in a packages registry. Github offers this already, but we're not using that yet, so you will need to build the docker images manually. You can use `docker-compose build` to do so.

Prerequisites:
1. install Docker-for-desktop on windows
2. Enable Kubernetes in Docker-for-desktop in the settings
3. Have `kubectl` installed, ready from the command line.
4. Make sure you kubectl context is pointing to your local kubernetes environment. If you have no other k8s clusters, that should already be the case. 

How to run:
1. Edit your host file as mentioned above. Add the following line in your `C:\Windows\System32\drivers\etc\hosts` file: `<Your IP Address> identityadmin.k8s.local`
2. Run the following commands
```
cd k8s
kubectl apply -f ./namespace.yml
kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.3/manifests/metallb.yaml
kubectl apply -f ./metallb.yml
kubectl apply -f ./sqlserver.yml
kubectl apply -f ./identity.admin.yml
kubectl apply -f ./identity.sts.yml
kubectl apply -f ./identity.api.yml
```    

I might have missed a couple of things, so if someone could test this out on their machine, that'd be great.

### DIFF
--- a/k8s/identity.admin.yml
+++ b/k8s/identity.admin.yml
@@ -1,0 +1,100 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    system: identityadmin-demo
+    app: identity-admin
+    version: "1.0"
+  name: identity-admin
+  namespace: identityadmin-demo
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      system: identityadmin-demo
+      app: identity-admin
+  template:
+    metadata:
+      labels:
+        system: identityadmin-demo
+        app: identity-admin
+        version: "1.0"
+    spec:
+      containers:
+
+      - env:
+        - name: ASPNETCORE_ENVIRONMENT
+          value: Development
+        - name: ConnectionStrings__ConfigurationDbConnection
+          value: Server=sqlserver;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true
+        - name: ConnectionStrings__PersistedGrantDbConnection
+          value: Server=sqlserver;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true
+        - name: ConnectionStrings__IdentityDbConnection
+          value: Server=sqlserver;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true
+        - name: ConnectionStrings__AdminLogDbConnection
+          value: Server=sqlserver;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true
+        - name: ConnectionStrings__AdminAuditLogDbConnection
+          value: Server=sqlserver;Database=IdentityServer4Admin;User Id=sa;Password=Password_123;MultipleActiveResultSets=true
+        - name: AdminConfiguration__IdentityAdminBaseUrl
+          value: http://identityadmin.k8s.local:9000
+        - name: AdminConfiguration__IdentityAdminRedirectUri
+          value: http://identityadmin.k8s.local:9000/signin-oidc
+        - name: AdminConfiguration__IdentityServerBaseUrl
+          value: http://identityadmin.k8s.local
+        - name: AdminConfiguration__RequireHttpsMetadata
+          value: "false"
+        - name: IdentityServerData__Clients__0__ClientUri
+          value: http://identityadmin.k8s.local:9000
+        - name: IdentityServerData__Clients__0__RedirectUris__0
+          value: http://identityadmin.k8s.local:9000/signin-oidc
+        - name: IdentityServerData__Clients__0__FrontChannelLogoutUri
+          value: http://identityadmin.k8s.local:9000/signin-oidc
+        - name: IdentityServerData__Clients__0__PostLogoutRedirectUris__0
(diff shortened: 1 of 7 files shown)


skoruba:
Thanks @bravecobra - we will test this PR with @TomasHubelbauer

ekjuanrejon:
@bravecobra 

I think a helm chart would be better for Kubernetes. I have done so in a private repo that unfortunately can't share. I will soon create helm charts in my own fork

bravecobra:
@ekjuanrejon,
I agree, this first PR is just a proof that we can have a working environment on k8s. Putting such into a helm chart would indeed be the next step. The current images however could use some extra tcl before we can put them into production. The seed command for instance doesn't survive a subsequent run. That seed should run as an init-container.

skoruba:
@bravecobra - what do you suggest regarding to seed? Is it necessary to rework seed in project or event extract seed into separate project like console app for instance?

bravecobra:
There are 2 options there: either 

1. we let the seed do nothing if records already exist in the database (now it crashes on that) and assume thus assume a previous seed already did the job
1. use the /seed to do only the seeding and don't start listening so we can use this command-line option to only do the seeding part and nothing else. That way it can be used as an init-container.

I don't think a separate console project is required.

skoruba:
First use case is already covered, this should not crash it - before populating of data is check if something already exist in table, otherwise do nothing.

I faced another issue which is not solved yet, sometimes database is not ready and because of this seed failed.
What do you suggest with this issue?

@xmichaelx suggested library called Polly (https://github.com/App-vNext/Polly) - and add some retry mechanism with delay.

What do you think?

ekjuanrejon:
Actually in production, I would strongly recommend to use a SQL managed service from a cloud provider

Am currently using manged SQL service from Azure. I think that is the best scenario as the backing and restoring is managed by the cloud provider. No need to mound disk into kubernetes etc...

@skoruba when running integration tests or running locally via docker-compose it is good to use 
`restart: always`

https://github.com/ufoscout/docker-compose-wait

or write a bash script... 

Have done a combination of those

bravecobra:
I agree, you would never run a sql database server in kubernetes in production. Hence the message in the beginning of the readme. Most cloud providers have their own/better solutions for that, that scale better you would ever be able to achieve.
However, again, for development purposes,  this would work for 99% of the cases.

bravecobra:
Ideally we would have a helm chart that would conditionally include the correct stable helm chart of the configured database

Luckily those are available already:

* https://github.com/helm/charts/tree/master/stable/mssql-linux
* https://github.com/helm/charts/blob/master/stable/mysql
* https://github.com/helm/charts/blob/master/stable/postgresql

Conditionally including a sub-chart is possible (https://stackoverflow.com/questions/54032974/helm-conditionally-install-subchart), however we need to keep https://github.com/helm/helm/issues/3742 in mind.

We also need to keep metallb out of the chart, as that won't be required for cloud providers. I'm only included it for completeness if you want to try this out on a bare metal cluster.

Although the API part is currently optional, I'd like to include that always. Personally I find the Admin is to call an API instead having the domain logic built-in as we currently have, but that a topic for another issue. Another thing I still need to test out is how the setup responds when Istio injection is enabled on the namespace as well as log-shipping to seq for instance. I first want to test these things out before pouring it into a more definitive solution like a helm chart. If we were to make a helm chart, I'd propose a separate PR that could handle production situations when deploying, not that dev test thingie.

Luckily I prefixed this PR with WIP (Work in progress)  😄

ekjuanrejon:
@bravecobra 

How about the following:

1.  Use a config map or secret to avoid loads or environment variables, or repetitive environment variables
2) For the k8s services it it possible to use an ingress instead of a load balancer,

bravecobra:
@ekjuanrejon sure, a config map would be more k8s solution. See the missing features in the description where I already mention that. I just used the environment variables to keep the analogy with the docker-compose setup we already had. 
An ingress would be easy way out, however then we wouldn't be able to test the statelessness of the current images. How would the service behave if we start adding replicas. In other words, I have no idea at this point whether scaling would work for the current images, just something I wanted to have tested out.

ekjuanrejon:
@bravecobra if that is your feature then you can scale your replicas to 1 and still use the ingress controller. it is better to use the ingress controller than to create three load balancers in the cloud

bravecobra:
It's not a matter of fear, it's about scalability and being able to test it.

PR #438, opened 2019-12-11
Expose IMvcBuilder from AddMvcWithLocalization  back to ConfigureServices

### LINKED ISSUE #437: Unable to cleanly add additional MVC configuration in ConfigureServices
Hi,

I have need to add additional controllers and views to STS.Identity but since ASP.NET Core support controllers and views in separate library I went that route. I have created new library, added my controllers and views and created an extension method to register everything in STS.Identity but  `IMvcBuilder`is not available in ConfigureServices.

Usualy it's normal that `IMvcBuilder` is available in ConfigureServices since `AddConttrollersWithViews` is called in ConfigureServices. 

```csharp
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddControllersWithViews();
        }
```

But, since it's moved to StartupHelpers.AddMvcWithLocalization and that method returns void it's not possible to change just Startup.ConfigureServices and add extension method.

I will create PR that fixes that so there is no need to change StartupHelpers to get to IMvcBuilder.

b0:
Since most of extension methods needed to configure MVC operate on `IMvcBuilder` exposed by `AddControllersWithViews` I find it missing in ConfigureServices because custom `AddMvcWithLocalization `doesn't return it.

This PR changes this so if additional configuration changes is needed there is no need to change Skoruba's StartupHelper.cs but Startup.cs ConfigureServices.
Fixes #437


skoruba:
Thanks @b0 👍🏼🙂

PR #444, opened 2019-12-18
Expose IIdentityServerBuilder in ConfigureServices

b0:
Hi @skoruba ,

as was with `IMVCBuilder`, `IIdentityServer` isn't available in `ConfigureServices` so it's impossible to further configure `IdentityServer4` without modifying `StartupHelpers` private methods.

This pull request made `AddIdentityServer` method public and return builder back to caller.


skoruba:
It looks good, thanks 👍

PR #447, opened 2019-12-27
Logging to the console

### LINKED ISSUE #435: Add the option to log to the console
Add the sink to let Serilog log to the console: `Serilog.Sinks.Console`. This is especially useful in environments like Kubernetes where you want to do logshipping/distributed logging. It can be off by default, although that would not be my preference. 

The recently added logging to the DB is definitely not my preferred way to keep logs. You lose the option to do structured logging as well. I would vote to make this optional as well. I want to be able to do log-shipping to off-site services like Seq/Cloudwatch/Datadog. Currently there is no way to scrape those logs, except from file, but that requires write access to the file-system which in a Docker environment isn't desirable either.
We also need to revisit the current way of logging to enable structured logging so that services like seq and cloudwatch can do structured queries on certain properties. Adding distributed tracing with OpenTracing/Jaeger is certainly another thing we need to take a look at.

bravecobra:
Since I needed the console logging for k8s, I went ahead and implemented the console logging as agreed upon in #435 .
Missing is the documentation for logging to the database as that was removed (Serilog.Sinks.MSSqlserver is still being referenced, so it's only a matter of adding the configuration.) I also didn't remove the error logs UI, but that'll remain empty with the current config (only console and debug for VS).
Meanwhile I fixed a little Admin.API issue for Docker in the swagger UI that slipped in somewhere as there is an ApiBaseUrl config param now that didn't get overridden yet.

Closes #435


skoruba:
Great, thanks 👍🏼

skoruba:
@bravecobra - I've used almost all your changes and manually add into dev branch.
I've excluded AppSettings.Development.json files.
Thanks 👍

PR #452, opened 2020-01-01
Feature/appsettings refactor

ekjuanrejon:
- Reduce complexity of appsettings.json
- Move serilog to its own json file.
- move identitydata to own json file
- move identityserver to own json file

In so doing it will allow for configuration to mounted from the outside into the container. This allows anybody to customize serilog and the related identity data without having to use loads of environment variables.

Identity data could have been more granular, example a json file only for clients.

When deploying with kubernetes it is easier as those json files can be deployed as configmaps or secrets and then mounted to the container


skoruba:
@ekjuanrejon - this is perfect idea, thanks!

bravecobra:
@skoruba you did undo the serilog changes, where we decided to only include the console and leave the rest optional. The current dev now has the mssql and file sinks included again.

skoruba:
@bravecobra - yeah, thanks, I will revert it.

ekjuanrejon:
@bravecobra how about mounting the serilog.json from host to container, this way the config can be different per developer

bravecobra:
@ekjuanrejon sure you could do that perfectly. It's more a question what is to be the default. Logging to the database is highly opinionated way of doing that. I would never want logging to affect the performance of my database in any way. Setting logging to be verbose (like Debug of Trace) would affect the latency. It would never want that to be the default in any way.

bravecobra:
I also noticed the serilog proposed way of handling exceptions in the Main method as removed as well. If a docker container would crash due to an unhandled exception, you would never find the entry in the logs, your first resource to start to investigate the problem. I find that concerning. Please add that back.

ekjuanrejon:
@bravecobra I kind of disagree with you on that. I can be totally wrong.

I get your point that serilog will not log the unhandled exception. It is my understanding in the container world that the environment is in charge of log aggregation etc... I am of the opinion that if it can be delegated to the environment then it's better. This way you can have many log collector agents and transform the data as needed.

So basically do console logging and let some other process collect the logs. For example fluentbit, fluentd and then send it to ELK. So once the exception is on the console the log collection agent will pick it up and forward.. 

We currently use datadog

I can change it...don't have a problem

bravecobra:
Sure, the difference is who/what writes to the console, **but** there is also security concern here as well. Any application that throws unhandled exceptions is subject to attack. If you know the memory address at which the app "crashes" (because that's what it is), a hacker can make sure that there a payload ready to be executed at the memory address. That allows a hacker to gain arbitrary access with the container.
In general, never let your application throw unhandled exceptions if you don't want your app to be vulnerable to hacking.

PR #453, opened 2020-01-04
Catch any unhandled exception, log and exit cleanly

bravecobra:
As I mentioned in #452, letting any unhandled exception leak out of the program and thus when it "crashes", is a security vulnerability as it allows a hacker to execute arbitrary code when the program crashes. This PR should fix that.


bravecobra:
The failing build seems to be a race condition.

PR #466, opened 2020-01-22
Fix/missing culture select

aiscrim:
When only one culture is configured, the cultureSelect is not shown. This was causing an error in the language.js script that this PR fixes.


skoruba:
Thanks, this file is also in admin ui probable. 😊 I will check it and fix it. 👍🏼

PR #475, opened 2020-01-28
Find users having some Claim by Admin.Api

b0:
Hi @skoruba 

currently it's not possible to find users having some claim by using API. 

Let's say i have 200 users having claim "EyeColor":"Blue". Currently you can find them only by getting all 200 users and then asking for claims for every user. That's 200 requests.

This PR creates new controller `UserClaimsController`, new method on repository, service, new audit event and make it possible to search by ClaimType and ClaimValue.

Best regards


b0:
Hi @skoruba ,

what is your view of this PR? Do you find it ok for inclusion in Administration api?

Thank you

skoruba:
Can we move this code from controller to UsersController?

b0:
I was thinking about it but wasn't sure of route...

Do you have idea?

skoruba:
I would create method called:

GetUsersByClaim
[HttpGet("ClaimType{claimType}/ClaimValue/{claimValue}")

What do you think?

Thanks for your great work. 👍🏼

Issue #245, opened 2019-04-07
Add German (de) Localization


schulz3000:
Add German (de) Localization.

If you want I can provide a PR for it.

skoruba:
It would be great. 👍 
Please, send a PR.
Thanks!

skoruba:
Any progress on this? :)

PR #480, opened 2020-01-30
adding 🇩🇪 tranlation

thisIsTheFoxe:
### Adding a German translation. #245 

I hope i got everything... ^^
Lmk if anything's missing..


skoruba:
Thanks @thisIsTheFoxe - I will check, I really look forward to german translation. 👍

skoruba:
@thisIsTheFoxe - It looks perfect - one thing which I mentioned - you've used `exisiert` insted of `existiert`. Is it correct? I do not know german, this says google translate. :-)
Thanks.

thisIsTheFoxe:
Wow. Good catch.. Translations are a typo feast.. I bet that's not the last one, but with that fixed it should be down to a minimum (hopefully).. ^^

PR #484, opened 2020-02-05
Add ability to specify a From email address separate from the login username

jwilliamsnephos:
Some SMTP services require a unique username.  The existing code assumes that the login is also the "from" e-mail.  My specific use case is when using AWS SES for smtp, the login username is not valid to be used as the from e-mail address.  The login isn't an e-mail address at all.

This provides backwards compatibility, by introducing the "From" as an optional configuration that can be supplied by the user.

If there are other places where this modification needs to be considered, please let me know, I'll be happy to do the work.  If you would prefer my implementation to be different, I'd be happy to refactor accordingly.


skoruba:
Hi @jwilliamsnephos - it looks good, thanks for sending this.

PR #499, opened 2020-02-14
check for existence or role and claim before doing operation

b0:
Hi @skoruba,

currently when you try to PostUserClaims or PostUserRoles with non existent roleId, claimId api would break with NullReferenceExcpetion and return 500.

I just added missing check so now it returns ProblemDetails and there is no exception.

Thank you


b0:
Hi Jan,

did you checked out two added lines? I still thin that check should be there so we return ProblemDetails instead of exception...

Thank you

skoruba:
Hey @b0 - thanks for this PR, it looks great. 👍🏼

PR #508, opened 2020-02-20
Fix CORS issue when not launching with iis

sheng-jie:
When using Skoruba.IdentityServer4.Admin.Api launch setting, you will get the error as below.
```
Fetch errorFailed to fetch http://localhost:5001/swagger/v1/swagger.json
Fetch errorPossible cross-origin (CORS) issue? The URL origin (http://localhost:5001) does not match the page (http://localhost:5002). Check the server returns the correct 'Access-Control-Allow-*' headers.
```


skoruba:
Hi @sheng-jie - thanks for this PR! :-)

PR #509, opened 2020-02-20
move port and environment config to override.yml

b0:
Hi @skoruba ,

I have made small changes to docker-compose.yml and docker-compose.override.yml.
Currenty, they were incorrectly used. 

docker-compose.yml should define services but for example not ports nor some of environment variables.

Ports should be moved into .override.yml for development purposes, and another, new file for production or staging.

Great and simple example is on [Docker compose](https://docs.docker.com/compose/extends/) and it deals with this exact situation: define services, and in overrides ports and envirnoment(development, production)

Anyway, this PR will not change how docker containers behave, it's just done in a slightly better way.

Thank you

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -5,14 +5,15 @@ services:
     environment:
       - ASPNETCORE_ENVIRONMENT=Development
     ports:
-      - "80"
+      - 9000:80
     volumes:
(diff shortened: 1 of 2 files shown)


skoruba:
HI @b0 

thanks for this PR. 👍

Issue #258, opened 2019-04-19
Adding public origin to configuration of STS


xmichaelx:
In many cases STS is behind some proxy or load balancing so the default origin used in discovery document for prepending endpoints inferred from request may be a local one inferred from passed request.

[PublicOrigin](http://docs.identityserver.io/en/latest/reference/options.html) is an option that can be set to mitigate that, we should add it to the configuration of STS.

aiscrim:
I am developing a PR to deal with this and other issues related to load balancing configurations. Stay tuned! :)

skoruba:
Thanks @aiscrim

skoruba:
Done in master. Thx

PR #492, opened 2020-02-11
Invite user

jwilliamsnephos:
Adds the ability to invite a user into a system when public registration is not enabled.

See issue
https://github.com/skoruba/IdentityServer4.Admin/issues/174

PR #510, opened 2020-02-24
Feature/load balancing support

aiscrim:
In production scenarios where multiple Identity Servers lay behind a load balancer or a proxy, it may be necessary to configure a couple of settings in order to have everything working correctly:
- Data Protection: by default, the AntiForgeryTokens are signed using the local machine DPApi. As a consequence, if a form is loaded by one server and then posted to another server, the anti forgery token will not be considered valid. In this PR we configure the Identity Server to use a central database instead of the local machine to store these tokens.
- Public Origin: when the clients ask for the discovery document, by default all the endpoints are based on the Identity Server URI. When we have multiple servers, we may want to show the external facing URI instead. This is achieved by setting the PublicOrigin option, that we are now exposing in the appsettings under a new AdvancedConfiguration section.

Addresses issue #258.


aiscrim:
@sheng-jie I don't think it would be a good idea. The PublicOrigin setting is not a URL, but an origin (as in CORS origins). For example, if my IdS is hosted on a server called ids.mycompany.com and on a folder called IdentityServer, the IdentityServerBaseUrl would be https://ids.mycompany.com/IdentityServer while the PublicOrigin would be https://ids.mycompany.com

Using the same name for settings with different meanings is not a recommended practice, so I'd leave it as it is.

b0:
Hi, @aiscrim, @skoruba  

I still didn't had time to try so I'll fist ask,
do you know if with this included in AdminAPI/AdminUI project generating let's say password reset token(token provider under the hood uses DataProtection) will work same as if token is generated in STS? 

Thank you

aiscrim:
Hi @b0,
to my knowledge, configuring the data protection just changes the default location where the keys are stored, so it shouldn't impact any functionality.
Indeed, I tried the password change from AdminUI and it works as before, no issues there.

The only problem would be if you generated a password reset token in STS (forgot password functionality) before this change and the user tries to reset the password after the change. In this case the keys would be different, so it would probably fail, but I think we can ignore this transient problem.

b0:
Hi @aiscrim,
sorry, i wasn't clear enough. 

Password change from AdminUI works because _GenerateToken_ and _ResetPassword_ (use genarated token) are 2 lines apart in [here](https://github.com/skoruba/IdentityServer4.Admin/blob/be6fc55a61d890378dd21f9110c2850e963a487f/src/Skoruba.IdentityServer4.Admin.EntityFramework.Identity/Repositories/IdentityRepository.cs#L351-L353) so token will always be valid.

Whats bothering me is this situation(for now just hypothetical because this functionality still doesn't exists): 
let's say you want to implement "Forgotten password functionality" that already exists in STS on Login form but you want it in AdminUI. 
So you as admin have ability  to send user a reset password link to STS where she can reset her password on already available page. I am not sure the token/code generated in AdminUI (and sent via email) will be valid in STS.

Or the currently proposed PR #492  for inviting user.  If this is to be implemented, we should generate token on AdminUI, create "Set password" page on STS and validate token generated on AdminUI in STS project...

Hope, I am a bit clearer now?

Thank you

aiscrim:
Ok, I didn't understand fully your concern, sorry, now it's clear.
You are right, a token generated by AdminUI wouldn't work in STS and viceversa. What needs to be done is setting the same application name in all the applications that need to share tokens. I've already tested it locally and it works, I will create a patch asap.
Thanks a lot for your insight!

b0:
Np. We are all here to help as we can. Thank you for input. Great work.

Best regards

PR #528, opened 2020-03-09
Use correct logger instance for SmtpEmailSender

b0:
Small fix for SmtpEmailSender to use correct ILogger

Thank you


skoruba:
Thank you 👍🏼

PR #530, opened 2020-03-09
Add ResourcePath config to Admin.Api

b0:
As Admin and STS projects use Resources from Resources folder Api project should use that too.
IGenericControllerLocalizer is already in use but fails bcs config is missing.

This PR fixes that.

Thank you


b0:
Just a question, this is approved but not merged... how so? :)

Thank you

skoruba:
Done :-)

PR #543, opened 2020-03-14
Update Danish translations for Identity (STS) FE

carlreid:
I plan on using this solution for a Danish audience and just ran through the translations with my Danish girlfriend. She's non-technical too, so this would best represent a non-technical audience. So thought I would contribute these changes back.

This doesn't include any changes to the Admin side but that's not so important in my use case and a few translation issues probably doesn't matter so much there.


skoruba:
Thanks 👍🏼

PR #548, opened 2020-03-24
Set common application name for data protection between Admin and STS.

aiscrim:
Fixes #510


skoruba:
What about API project? :)

aiscrim:
The Api doesn’t generate AntiForgery tokens nor reset tokens (if I’m not missing something), so it doesn’t use anything from the DataProtection Api.
Regarding the application name, let me know your decision and I can change it to whatever you think appropriate.

b0:
I think API should have i too. 
I am pretty sure in one moment we will try to rewrite STS and Admin to use same API because now we have code duplication and inconsistencies. Eg. I would like to initiate "forgotten password" flow from my own app using API... Sure, i would need to add that functionality but I would still need to provide link to STS form for setting password and STS should validate token. 

If application name for data protection is not same on all projects - this will not work... :( 

Best regards

Issue #549, opened 2020-03-25
Serilog setup does not include environment specific configuration files


ViRuSTriNiTy:
Hi there,

i would like to add environment specific configuration files like `serilog.Development.json` or `serilog.Staging.json`. Currently only `serilog.json` is added hence environment specific configuration files cannot be used.

I can provide a PR. What do you say? 

So lonG
Daniel

skoruba:
Hi, please check dev branch, I added this.
Thanks

ViRuSTriNiTy:
Hi @skoruba, thx for the fast reply. I cannot find the code that adds this functionality. Would you please provide a link to relevant commit? So lonG

skoruba:
Sorry, I added support for loading configuration from secrets. I am wondering about this one, I tought that this is build in default configuration, did you test it?
Thanks!

skoruba:
I added this: https://github.com/skoruba/IdentityServer4.Admin/commit/76a0181702bafedc03f6004ccb8ad6f5dbae2e45

b0:
Hi @skoruba,

I was thinking the same way as you. Since AddJsonFile magicaly loads appstettings.{environment}.json I was thinking that is true for all files - but it is not.  [configuration](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#default-configuration)

`
            .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
            .AddJsonFile("serilog.json", optional: true, reloadOnChange: true)

            .AddJsonFile($"serilog.{environment}.json", optional: true, reloadOnChange: true)`

Sorry for bad formating. I was have hard time with it 😠

ViRuSTriNiTy:
@b0, @skoruba 

AddJsonFile() does **not** load appstettings.{environment}.json magically. This is a implementation detail in the DefaultBuilder, see https://github.com/dotnet/aspnetcore/blob/master/src/DefaultBuilder/src/WebHost.cs#L170

My PR would add some lines according to the WebHost implementation like

    configApp.AddJsonFile("serilog.json", optional: true, reloadOnChange: true);

    // https://github.com/skoruba/IdentityServer4.Admin/issues/549
    //
    // add configuration files based on hosting environment,
    // adopted from https://github.com/dotnet/aspnetcore/blob/master/src/DefaultBuilder/src/WebHost.cs#L170
    var env = hostContext.HostingEnvironment;

[...]

PR #551, opened 2020-03-25
- IdentityServer4, Admin, Admin.Api, STS.Identity, added support for environment-specific configuration files

ViRuSTriNiTy:
… in logging setup and `CreateHostBuilder()`, see issue #549


src/Skoruba.IdentityServer4.Admin.Api/Program.cs
```diff
@@ -38,7 +38,9 @@ private static IConfiguration GetConfiguration(string[] args)
             var configurationBuilder = new ConfigurationBuilder()
                 .SetBasePath(Directory.GetCurrentDirectory())
                 .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
-                .AddJsonFile("serilog.json", optional: true, reloadOnChange: true);
+                .AddJsonFile($"appsettings.{environment}.json", optional: true, reloadOnChange: true)
```
b0: Do we really need this line ? I think apps already works with appsettings{environemnt}.json by default.

src/Skoruba.IdentityServer4.Admin/Program.cs
```diff
@@ -61,7 +61,9 @@ private static IConfiguration GetConfiguration(string[] args)
             var configurationBuilder = new ConfigurationBuilder()
                 .SetBasePath(Directory.GetCurrentDirectory())
                 .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
-                .AddJsonFile("serilog.json", optional: true, reloadOnChange: true);
+                .AddJsonFile($"appsettings.{environment}.json", optional: true, reloadOnChange: true)
```
b0: Do we really need this line ? I think apps already works with appsettings{environemnt}.json by default.

src/Skoruba.IdentityServer4.Admin.Api/Program.cs
ViRuSTriNiTy: Imho yes, this line is necessary as there is no "automagic" that is going on here, see https://github.com/skoruba/IdentityServer4.Admin/issues/549#issuecomment-603730495

src/Skoruba.IdentityServer4.Admin/Program.cs
ViRuSTriNiTy: Imho yes, see https://github.com/skoruba/IdentityServer4.Admin/pull/551/files#r397750808

b0:
At least how I am reading this: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#default-configuration

appsettings.json and appsettings.{environment}.json are "magicaly" provided by default. 

Best regards

ViRuSTriNiTy:
@bo Yes i read this section too but i think its a bit misleading. I've checked the .NET Core sources and as i mentioned in the related issue the `appsettings.{environment}.json` is added by the `DefaultBuilder`, see https://github.com/dotnet/aspnetcore/blob/master/src/DefaultBuilder/src/WebHost.cs#L170. The thing here is that the logging configuration is created without the `DefaultBuilder` hence you need to do it manually.

b0:
Hi @ViRuSTriNiTy ,

indeed, logging configuration is created differently and those lines are needed. 

Great work. 👍 

Thank you

ViRuSTriNiTy:
@b0 Thanks for reviewing. How to remove the merge block? Is @skoruba needed to resolve this?

ViRuSTriNiTy:
@skoruba Please unblock the PR when everything is ok.

src/Skoruba.IdentityServer4.Admin.Api/Program.cs
ekjuanrejon: I don't think we need this line. If anyone wants a custom serilog configuration. The configuration should be mounted into the application. 

Define in your release pipeline to mount the serilog.json from the outside to the inside. The proposed line will not work with docker and kubernetes

ViRuSTriNiTy: > I don't think we need this line. If anyone wants a custom serilog configuration. The configuration should be mounted into the application.
> 
> Define in your release pipeline to mount the serilog.json from the outside to the inside. The proposed line will not work with docker and kubernetes

Thanks for your feedback but i don't understand your instructions. What does "mounted into the application" actually mean and what "release pipeline"? Do you mean i should copy Serilog.Development.json to Serilog.json before releasing the application?

ekjuanrejon: if you look at the docker compose below
https://github.com/skoruba/IdentityServer4.Admin/blob/master/docker-compose.yml#L35

This give the flexibility for anyone to define serilog.json as they want and mount it from the host machine to the container

ekjuanrejon: In docker or kubernetes if you have multiple applications that required serilog you only want to have one config. In kubernetes you can use a config map or secret and mount those inside the pod. 

If it need to change you change the config map without having to redploy your application

ViRuSTriNiTy: @ekjuanrejon So this PR is not necessary at all?

ekjuanrejon: To me it is not necessary but if you want I'd say if you still want to do it. do it in your fork

ViRuSTriNiTy: So you want only "necessary" changes not improvements, that's sad as my approach also [allows config transformations](https://stackoverflow.com/questions/53208920/equivalent-to-app-config-transforms-for-net-core) via [SlowCheetah](https://marketplace.visualstudio.com/items?itemName=vscps.SlowCheetah-XMLTransforms), all effort in vain. Closing the PR.

ViRuSTriNiTy:
PR closed as a result of https://github.com/skoruba/IdentityServer4.Admin/pull/551#discussion_r407981004

src/Skoruba.IdentityServer4.Admin.Api/Program.cs
ekjuanrejon: I am not the owner of the repo. I just gave a comment.

ekjuanrejon: The owner can very well decide to approve your pr

ekjuanrejon: @skoruba

skoruba:
@ViRuSTriNiTy - please, reopen this PR. I want to use it.
Thanks.

ViRuSTriNiTy:
@skoruba Reopened as requested

skoruba:
Thanks 👍🏼

PR #558, opened 2020-03-30
return `UserName` from external login provider

WeihanLi:
return `UserName` from external login provider so that user could create account without refill 
the `UserName` field


skoruba:
Thanks 👍

PR #564, opened 2020-04-01
Improved suggested claims.

inbarbarkai:
Hi,

I was adding a lot of claims lately and it was very annoying that it wasn't suggesting me the claim type  I entered just a second ago.

I've added logic to get the claims that are required by existing API resources.

*Thinking points:*
1. That logic might be out of place and should be somewhere a little higher in the food chain.
2. Maybe it should also prioritize by actual usage.
3. Should it have a list in the database that is manageable?

### DIFF
--- a/src/Skoruba.IdentityServer4.Admin.BusinessLogic/Services/ClientService.cs
+++ b/src/Skoruba.IdentityServer4.Admin.BusinessLogic/Services/ClientService.cs
@@ -304,9 +304,9 @@ namespace Skoruba.IdentityServer4.Admin.BusinessLogic.Services
             return protocolTypes;
         }
 
-        public virtual List<string> GetStandardClaims(string claim, int limit = 0)
+        public virtual async Task<List<string>> GetStandardClaimsAsync(string claim, int limit = 0)
         {
-            var standardClaims = ClientRepository.GetStandardClaims(claim, limit);
+            var standardClaims = await ClientRepository.GetStandardClaimsAsync(claim, limit);
 
             return standardClaims;
         }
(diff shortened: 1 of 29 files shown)


skoruba:
Hi,
thanks for this PR, I will test it. This is great idea, maybe some database table for this will be good, I will think about it.

inbarbarkai:
I can't rewrite the request with a backing table if you want.

skoruba:
Great idea, I am thinking about it. Do you have time create some UI for managing theses claims as well or first step is put these claims into database? 
Thanks

inbarbarkai:
I'll do a backing table first with some seed.
UI is not really my thing so we'll call it step 2.
I'll give it a shot but I can't make promises on the end result.

Do you think this API should be moved elsewhere?

skoruba:
OK, good, you can please add this new table and I will add UI sometimes later, this is good for starting point. 😊

inbarbarkai:
Implemented backing table with seed.
I've added infra for future sort by usage and last use.
I want to add an enum for that sorting purpose.
Where is it appropriate to place it? I couldn't decided based on existing files.

skoruba:
What do you plan to put into this enum - ASC/DESC?
Can you please generate extra new migration for this new table? I think this is mixed in DbInit migrations. ;-)
Thanks - great work, I really appreciate your help 👍

inbarbarkai:
I want to add enum for ASC/DESC.
I also consider a way for specific on which column to sort (maybe a string).

> Can you please generate extra new migration for this new table? I think this is mixed in DbInit migrations. 

You mean an repo and context as well? A new migration sub-folder for standard scopes?

skoruba:
I mean migration name, currently is generated under DbInit migration name. What do you thing? Btw: what about boolean - sortAsc?

inbarbarkai:
A boolean `sortAsc` sounds great, I'll implement it tomorrow as well as the separate migration.

inbarbarkai:
@skoruba I've implemented what we discussed.

skoruba:
Thank you 👍

PR #567, opened 2020-04-07
update Dockerfile for alpine based image

### LINKED ISSUE #497: Consider an Alpine dockerfile
The normal aspnet-buster is all good, for sure.  We've converted to using aspnet-alpine for our stuff, and it reduces the size of the docker image.

WeihanLi:
update `Dockerfile` for alpine based image fix #497


WeihanLi:
CI Error is as follows, seemed like a concurrent problem

>   System.IO.IOException : The process cannot access the file 'C:\projects\identityserver4-admin\tests\Skoruba.IdentityServer4.STS.Identity.IntegrationTests\bin\Debug\netcoreapp3.1\tempkey.rsa' because it is being used by another process.

PR #574, opened 2020-04-15
Few improvements when checking if a request is coming from a local subnet

aiscrim:
- checking for X-Forwarded-For header
- instead of scanning the network interfaces where the Identity Server is installed, we now check the request IP against a regular expression to identify private addresses (basically 192.169.*.*, 172.16-31.*.*, 10.*.*.* addresses will be considered private)
- added possibility to exclude in the configuration subnets from those that would normally be considered local
- added debug logging to help investigate why a request is considered local or not


skoruba:
Thank you 👍🏼😊

PR #609, opened 2020-05-14
Improvements to Forgot Password functionality

aiscrim:
- added possibility to reset password by username
- if the specified username or email does not exist, this method no longer reveals it


skoruba:
Thanks, let me check it. ;)

PR #628, opened 2020-06-17
post client secret route: add hash type parameter

klyse:
- added ignore for appsettings*.json to remove ignore development configuration
- added HashType parameter in client secret creation


skoruba:
Hi @klyse - thanks for your PR!

PR #632, opened 2020-06-20
Fix failed test - NullReferenceException at HashClientSharedSecret(ClientSecretsDto clientSecret)

martinussuherman:
I'm proposing this to simplify comparison of HashType. 

Hopefully this will also solve the failing test
```
[xUnit.net 00:00:04.72]     Skoruba.IdentityServer4.Admin.UnitTests.Services.ClientServiceTests.GetClientSecretAsync [FAIL]
  X Skoruba.IdentityServer4.Admin.UnitTests.Services.ClientServiceTests.GetClientSecretAsync [56ms]
  Error Message:
   System.NullReferenceException : Object reference not set to an instance of an object.
  Stack Trace:
     at Skoruba.IdentityServer4.Admin.BusinessLogic.Services.ClientService.HashClientSharedSecret(ClientSecretsDto clientSecret) in C:\projects\identityserver4-admin\src\Skoruba.IdentityServer4.Admin.BusinessLogic\Services\ClientService.cs:line 39
```

I'll create this as draft. Feel free to check.

Regards, Martinus

PR #637, opened 2020-06-22
Add base path for deployment in folder

martinussuherman:
This will allow for deployment in folder (https://www.example.com/stsidentity) for the token service and admin.
For token service, if behind reverse proxy, then need to set PublicOrigin to https://www.example.com too.

Regards, Martinus


martinussuherman:
Seems this PR will require #632 to be applied first so the checks will succeed.

skoruba:
Thank you for your changes. 👍

martinussuherman:
You're welcome 👍

PR #643, opened 2020-06-26
Draft: Fix for appveyor test fail silently (database related exception)

martinussuherman:
I gonna make this a draft, since this was meant primarily as proof of concept for the fix to the silently failing (false success) integration test.

A few things to note: 
- There are lots of commits, since this was mostly trial and error approach to find out the problem.
- If so desired, I can create a "cleaner" PR (more structured and less commits).
- Db config was read from  appsettings.json, I'm thinking of using other .json (appveyorsettings.json?) to store the test db config. Not sure how to do that though, any inputs are welcome :)
- I got my SSH public key there, because one of the builds get stuck and timed out. That would need to be change to template, or the SSH access might/should be removed.

Regards, Martinus

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,5 @@
+image: Ubuntu
+build:
+  project: Skoruba.IdentityServer4.Admin.sln
+  verbosity: minimal
+version: '1.0.{build}'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,10 @@
 image: Ubuntu
 
+environment:
+  APPVEYOR_SSH_KEY: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIO331U/KDT0FmRuw7pRLRkCCkceP1eURu4T4V+dsK4Ky ed25519-appveyor
+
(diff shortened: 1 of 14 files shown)


skoruba:
I do not know why db does not work properly in integration test, because db should be configure as memory database. Any ideas?
Thanks for your tips and PRz

martinussuherman:
> 
> 
> I do not know why db does not work properly in integration test, because db should be configure as memory database. Any ideas?
> Thanks for your tips and PR.

Hi @skoruba, thank you for the comment.

I hope I got your question right 😂

Well, I stumble upon [this article](https://docs.microsoft.com/en-us/ef/core/miscellaneous/testing/) a couple of weeks ago, that's the reason for my decision to use real db instead of in memory db for integration test. 

These are my arguments for doing so (from the above link):
1. In memory db is not relational db, so if we ever need to test for something that depends on relational constraints, then we shouldn't using it.
2. Integration test imo, should try to be as close as possible to reflect the deployment environment.

Of course, we might choose to agree to disagree, and that's fine 👌 

Regards, Martinus

PR #657, opened 2020-07-18
Docker support with HTTPS for local development

skoruba:


ekjuanrejon:
How to use it without https in local development?

skoruba:
Why without https?

ekjuanrejon:
I don't have a need to use https when doing local dev

ekjuanrejon:
Can I just comment the following line?
DockerHelpers.ApplyDockerConfiguration(configuration)

Issue #533, opened 2020-03-11
Add support for loading signing key from Azure Key Vault


skoruba:
Add support for loading signing key from Azure Key Vault 

Check this sample code:
https://github.com/damienbod/IdentityServer4AspNetCoreIdentityTemplate/issues/30

berendhaan:
@skoruba We have incorporated Azure Kv generated signing keys in our installation of your (magnificent) admin UI.
It it still rough and needs some more configuration options, but it could be a start for you. Shall I create a PR so you can have look?

skoruba:
@berendhaan - I would be perfect, could you please send a PR from dev branch?
Thanks!

berendhaan:
Done! #660 
I've cleaned some stuff up before I made the PR. Haven't tested this implementation but it should work since nothing major was changes.

PR #660, opened 2020-07-24
Add implementation for loading signing keys from Key Vault #533

berendhaan:
This feature can be enabled by setting `TokenSigningConfiguration.UseAzureKeyVault` to `true` and adding the following configuration options in the `appsettings.json`.

```
"AzureKeyVaultConfiguration": {
       "KeyVaultUri": "",
        "KeyName": ""
}
```

Some remarks:
- There is no caching of the keys
- Needs explanation how to configure Key Vault correctly
- Not battle-tested, at the moment max 3 users and signing in a couple of times a week


skoruba:
Hi @berendhaan - thank you for your PR, I've decided to load certificate from Azure KV on application startup currently - that is the reason why I am closing this PR. Thank you for your effort on this PR.

PR #726, opened 2020-10-06
Bugfix: Import All Windows User

sgebhardt:
The button to import all Windows Users from the AD is hidden when the In-Process hosting model is used.

If the in-proces hosting mode is used, IISServerOptions must be set. -> https://github.com/dotnet/AspNetCore.Docs/pull/9463#issue-228523235
So i added the options.

Added the missing partial view _StatusMessage.cshtml.


skoruba (approved):

skoruba:
Thanks 👍

PR #727, opened 2020-10-06
Features/hosting on azure doc

ioxFR:
Hi,

As in my case i use the solution without customizing the code, i would like to share my setup that can be really simply be reusable and ready to use solution.
- Using docker images published on docker hub (Admin & STS)
- Azure Managed Services (like app service linux container but also others like container instance)
- app service configuration to override appsettings.json properties and configure all you need in your contexte :)


skoruba:
Hi @ioxFR - do you plan to work in this? 😊👍🏼 Thanks

ioxFR:
Hi @skoruba 
Happy new year :)
Sorry for delay, yes i already started. I think i can finish them for end of january

skoruba:
Hello @ioxFR  - Happy new year 👍 that is geat news, I look forward to this PR. 😊

PR #730, opened 2020-10-09
Fix issues dealing with external providers and two factor auth.

zinkpad:


skoruba:
Thank you 👍 I will test it.

skoruba:
@zinkpad - I probable find another solution which works for me: https://github.com/skoruba/IdentityServer4.Admin/blob/dev/src/Skoruba.IdentityServer4.STS.Identity/Helpers/ApplicationSignInManager.cs#L65

I spent a few hours with testing of external providers and signout, especially with 2FA. And you were right, that this did not work correctly.

If I singed in via external OIDC provider, I got claim - "amr": "2fa" and "authenticationmethod":"oidc" and after that I got from IS4 claim called "idp":"local" which blocks correct signout from external provider.

Check this lines:
https://github.com/IdentityServer/IdentityServer4/blob/main/src/IdentityServer4/src/Hosting/IdentityServerAuthenticationService.cs#L125-L143

Now I pre-defined "idp" claim in SignInManager according to "authenticationmethod" claim value and it works.

What do you think?

zinkpad:
Hey @skoruba,

I tested it and works fine!!...after signin I get the claims {"idp": "oidc", "amr": "mfa"} and later signout works fine.

Also I noticed that you don't need to use the ExternalCookieAuthenticationScheme, I don't understand why in my previous tests I had to use it :( .....now works fine with ExternalScheme.

Next week I'll test it on my production environment...

Great job @skoruba !!

skoruba:
Great to hear from you 👍🏼👌🏻
Thank you for your help in this issue. 
I am closing this PR. 😊

PR #733, opened 2020-10-12
Feature/allow identity server data to be modified

ekjuanrejon:


skoruba:
What do you want to achieve? Initial seeding is designed that it is expected that you do not have any data in your DB.
Maybe could you please add some checks if data is already existing?

ekjuanrejon:
@skoruba 

> Maybe could you please add some checks if data is already existing?

That's what the code is doing. checking to see if the client etc.. exist or not...

My setup is the following. I have all clients as part of my ci cd process. So if I added a new client into env variables I want for the client to be inserted. Right now it does not insert... I have to manually script it...

You can also reject the pr and I'll keep the feature just on my fork

skoruba:
Oh, I did not notice that you have already checked this one. You can reopen it, no need to close. 😊
Thanks.

PR #746, opened 2020-10-23
Added support for portuguese language

labtrans-ufsc:
Hello, we from the STIGeo - LabTrans team at the Federal University of Santa Catarina - Brazil,  have added support for the Portuguese language so that users can better interact with Identity Server 4 with the Skoruba interface in our services.

We would like to make our small contribution to this incredible project that we are using and having a great experience.


skoruba:
Hi @labtrans-ufsc - thank you for your translation 👍🏼

skoruba (approved):

PR #759, opened 2020-12-02
Dutch translations added to resource (resx) files.

timmytimeless:
This contains a full set of Dutch translations in resx files. I've never done a pull request before so hope this works for reviewing and acceptance.


skoruba:
Thank you 👍🏼 Great work

PR #772, opened 2020-12-12
QR code white frame

vbjay:
- Force a white frame around qr code so that when the theme is changed to a dark theme, the code can still be scanned
- Include fonts in style build because they are needed by the styles
- Make it so running gulp builds all

After:
image

Before:
image


skoruba:
@vbjay - Thanks 👍

PR #775, opened 2020-12-15
Fixed Warning	CS0108: IAdminConfigurationDbContext.ApiScopes hides inherited member IConfigurationDbContext.ApiScopes

cognitive-arch:
Fixed Warning	CS0108	'IAdminConfigurationDbContext.ApiScopes' hides inherited member 'IConfigurationDbContext.ApiScopes'. Use the new keyword if hiding was intended.

I Removed the DbSet properties of ApiScopes and ClientCorsOrigin from the IAdminConfigurationDbContext interface because the same properties are available in the parent interface.


skoruba (approved):

skoruba:
@cognitive-arch - Thanks 👍👋

Issue #813, opened 2021-03-11
XSS issue in Client Secrets and Api Resource Secrets


skoruba:
In the views `ClientSecret` and `ApiResourceSecret` is not HTML encoded data attribute `data-secret-value` on the button:

```
<td><button class="secret-value-button btn btn-outline-primary" data-secret-value="clientSecret.Value"><i class="fa fa-eye"></i></button></td>
```

This data attribute is used in the dialog with secret detail.

Fix:
```
<button class="secret-value-button btn btn-outline-primary" data-secret-value="@Html.Encode(clientSecret.Value)"><i class="fa fa-eye"></i></button>
```

I will send this fix asap.

Thanks to Silton Santos for reporting.

skoruba:
Fixed

PR #814, opened 2021-03-15
Fix XSS in ApiSecrets and ClientSecrets #813

b0:


skoruba:
Thanks!

skoruba:
Fixed in 2.0.0., thanks!

PR #831, opened 2021-04-26
- fix identityserverdata.json in templates

pprometey:
error in 'templates\template-publish\content\shared\identityserverdata.json' file, **ApiScopes** is incorrectly defined


skoruba:
Thanks 👍

skoruba (approved):

Issue #832, opened 2021-04-26
Adding Client Token signing algorithms results in Error


IvoryWolf:
### Describe the bug
An error page is shown when trying to add additional token signing algorithms to a Client.
The selected items are not saved when reloading the client page. 

### To Reproduce

1. Create a client. 
2. Add any item from from the suggested items. I used ES256, ES384, ES512.
3. Save the client.

![afbeelding](https://user-images.githubusercontent.com/5112678/116084143-c0d20e00-a69d-11eb-82d9-284d5dcf8941.png)

### Relevant parts of the log file
```
{
  "TimeStamp": "2021-04-26T10:57:29.2887473",
  "Level": "Error",
  "Message": "Exception at route /Configuration/Client/3",
  "MessageTemplate": "Exception at route /Configuration/Client/3",
[...]

PR #833, opened 2021-04-26
Release/2.0.1

skoruba:
#832 - Adding Client Token signing algorithms results in Error


src/Skoruba.IdentityServer4.Admin.BusinessLogic.Identity/Skoruba.IdentityServer4.Admin.BusinessLogic.Identity.csproj
```diff
@@ -2,7 +2,7 @@
 
 	<PropertyGroup>
```
WeihanLi, line 3: Maybe we could manage these common properties in a `Directory.Build.props`, it'll be used for all the projects under this folder by MS Build implicitly,  I'd like to provide a PR if needed.

<https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets>

PR #843, opened 2021-05-02
Override OpenId redirect uri only when there's a value on configuration

### LINKED ISSUE #842: Only force the Redirect Uri on Admin.UI on OpenIdConnect Auth when explicitly defined on configuration
Currently we are modifying the Redirect Uri when authenticating using the `OpenIdConnect` provider with a value defined in configuration:

https://github.com/skoruba/IdentityServer4.Admin/blob/7023c0c6bb86f0d3c9bd2a5f603d17984c5c5fe3/src/Skoruba.IdentityServer4.Admin.UI/Helpers/StartupHelpers.cs#L420-L425

By default the `OpenIdConnect` provider is able to generate the callback URL based on the current hostname, port, etc:

https://github.com/dotnet/aspnetcore/blob/b7a2ec8c7ed6b48857af0a69688a73e8c14fe6cb/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L378

Now that we have the UI as a NuGet package, is not possible anymore to remove that behavior and let the OpenIdConnect provider generate the callback for you so you need to forcibly add that value.

We should only set that property if the user is trying to override the default, example:

```csharp
private static Task OnRedirectToIdentityProvider(RedirectContext context, AdminConfiguration adminConfiguration)
{
    if(!string.IsNullOrEmpty(adminConfiguration.IdentityAdminRedirectUri))
    {
        context.ProtocolMessage.RedirectUri = adminConfiguration.IdentityAdminRedirectUri;
    }

    return Task.FromResult(0);
}
```
What do you think @skoruba? I can open a quick PR with this if you agree.

Thanks!

cjaliaga:
With this change we allow to keep the default behavior unless the user explicitly tries to override the redirect uri if they have a custom configuration.

Fixes #842

### DIFF
--- a/src/Skoruba.IdentityServer4.Admin.UI/Helpers/StartupHelpers.cs
+++ b/src/Skoruba.IdentityServer4.Admin.UI/Helpers/StartupHelpers.cs
@@ -414,14 +414,17 @@ namespace Skoruba.IdentityServer4.Admin.UI.Helpers
             context.Properties.IsPersistent = true;
             context.Properties.ExpiresUtc = new DateTimeOffset(DateTime.Now.AddHours(adminConfiguration.IdentityAdminCookieExpiresUtcHours));
 
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
-        private static Task OnRedirectToIdentityProvider(RedirectContext n, AdminConfiguration adminConfiguration)
+        private static Task OnRedirectToIdentityProvider(RedirectContext context, AdminConfiguration adminConfiguration)
         {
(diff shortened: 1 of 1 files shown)


cjaliaga:
@skoruba thinking about defaults, I believe we should also either remove the `AdminConfiguration:IdentityAdminRedirectUri` from the Admin template's appsettings.json or at least set it as a blank string. Having the value there could make users believe that's something needed when it should be something for custom/specific scenarios.

Please let me know your thoughts on that :)

skoruba:
Thanks, I am sorry for delay. 😊

PR #854, opened 2021-05-14
Add Cors configuration to STS

ace90210:
This PR is to add support in the STS for Cors headers to be configured. This is required to resolve some errors with .Net 5 Blazor authentication which enforces a check on the Cors headers.

### DIFF
--- a/src/Skoruba.IdentityServer4.STS.Identity/Configuration/Constants/ConfigurationConsts.cs
+++ b/src/Skoruba.IdentityServer4.STS.Identity/Configuration/Constants/ConfigurationConsts.cs
@@ -21,5 +21,7 @@
         public const string AdvancedConfigurationKey = "AdvancedConfiguration";
 
         public const string CspTrustedDomainsKey = "CspTrustedDomains";
+
+        public const string CorsConfigurationKey = "CorsConfiguration";
     }
 }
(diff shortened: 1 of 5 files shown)


ace90210:
After future investigation i found the official method to handle Cors in the latest Identity Server. I have updated the PR accordingly. 

[Read here about official approach](https://github.com/IdentityServer/IdentityServer4/blob/9ef62e4b4c1e8e61043a7ba1f4ea474b6cb6a1b6/docs/topics/cors.rst)

LinQiaoPorco:
image

I assume that this is the same issue encountered.
Could those 2 commits solve the problem?
@ace90210

ace90210:
I believe so, i made this PR some months ago to solve the exact issue you posted and it resolved it for me at least.

LinQiaoPorco:
> I believe so, i made this PR some months ago to solve the exact issue you posted and it resolved it for me at least.

Thanks for replying!

skoruba:
When did you get this error?

LinQiaoPorco:
> 

Hi,  this error goes when I did not set-up Allowed Cors Origin for a client.
Then I set-up the Cors, it went well.

skoruba:
Hey @LinQiaoPorco - thanks for letting us know. I think the setup CORS for specific client is right way. :)

ace90210:
I can't recall if setting the client cors resolved the issue this was added for at the time but given it seems to not be an issue i will abandon this PR as it doesnt seem to be required

PR #871, opened 2021-06-09
Make correction towards the EF Core section

chungonion:
I noticed the database context name in the instruction is not correct. It should be `AdminAuditLogDbContext` instead of `AuditLoggingDbContext` 

Since not everyone is going to use that Powershell script to perform migrations, it will be great to rectify this part of the document. 

Thank you!


skoruba:
Thanks!

PR #896, opened 2021-08-14
Added option to exit after migrating the database

bravecobra:
I added an option to exit after applying the migration to the database. This allows using the Docker image to be used as an init-container and to do the database operations separately from the hosting part. This is especially useful in kubernetes deployments where you want to make sure everything is setup before starting any other containers that start listening.


skoruba:
Thanks! I am sorry for delay. 😊

PR #908, opened 2021-09-28
Fix Azure AAD external authentication not working since change to .net 5

ielcoro:
This change fixes skoruba#863 and most likely also skoruba#701 in the latests versiones. Since the change to Microsoft Identity library the Azure AAD provider does not work by default.

The problem relies in that Microsft Identity Library creates it's own cookie scheme handler, thus bypassing Identity Server (or asp.net identity) one when authenticating externally. Based on the info in the Identity Web Library repo, we can bypass this cookie handler by setting cookieScheme parameter to null.

This pull requests adds it to the default AAD external configuration provider with IdentityServer Admin.


skoruba:
Thanks!

skoruba:
Thank you again, I am sorry for delay. 😊

ielcoro:
You are welcome ☺️

Iñaki Elcoro
-----------------------------


________________________________

Issue #847, opened 2021-05-07
MySQL /seed trouble (#837)


aldinei-sampaio:
### Describe the bug
I had the same problem described at issue #837, but in MySql running on Windows.

I just want to share that I worked around the problem changing the MySQL migrations for the IdentityServerConfiguration DbContext. Just looked for "4000" inside the migration files and replaced it for "2147483647".

skoruba:
Thanks for reporting this, I will take a look at this. 😊

bravecobra:
As I was testing my helm charts for kubernetes adding support for other database, I noticed the same issue. The above trace is exactly the same. The used version is 2.0.1

bravecobra:
Any progress on this?

tbto:
It's may be Pomelo.EntityFrameworkCore.MySql bug, 
See: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1461

bitai-cs:
The problem is caused by the following line in the migration scripts: "longtext CHARACTER SET utf8mb4(4000)"
The data type 'longtext' should not be assigned a specific length.

I overcame the problem by removing the HasMaxLength (4000) from the migration scripts.

PR #931, opened 2021-11-21
fix mysql migrations error

nianwu:
[fix bug](https://github.com/skoruba/IdentityServer4.Admin/issues/847)


skoruba:
Thank you for this fix, did you try to generate new empty migration that this lines won't be added again?
Do you know how to avoid this issue?

nianwu:
I am not sure if regenerating the migration file will solve the problem
It seems to be caused by the MySQL nuget package

I just deleted the error code in the migration file

- google translate

PR #1030, opened 2023-01-13
- Fix for Timestamp Legacy Behaviour

alastor09:
- Fix for Postgresql complaining about DateTime not being UTC


alastor09:
Dear @skoruba please if you can review this request and merge it.

PR #1041, opened 2023-03-29
'Cannot write DateTime with Kind=Local to PostgreSQL' error fix

tuncaycemuzun:
The bug
I got the error below from Admin project. the endpoint was "https://localhost:44303/Configuration/Clients"
InvalidCastException: Cannot write DateTime with Kind=Local to PostgreSQL type 'timestamp with time zone', only UTC is supported. Note that it's not possible to mix DateTimes with different Kinds in an array/range. See the Npgsql.EnableLegacyTimestampBehavior AppContext switch to enable legacy behavior.

System.InvalidCastException: Cannot write DateTime with Kind=Local to PostgreSQL type 'timestamp with time zone', only UTC is supported. Note that it's not possible to mix DateTimes with different Kinds in an array/range. See the Npgsql.EnableLegacyTimestampBehavior AppContext switch to enable legacy behavior.
   at Npgsql.Internal.TypeHandlers.DateTimeHandlers.TimestampTzHandler.ValidateAndGetLength(DateTime value, NpgsqlParameter parameter)
   at Npgsql.Internal.TypeHandlers.DateTimeHandlers.TimestampTzHandler.ValidateObjectAndGetLength(Object value, NpgsqlLengthCache& lengthCache, NpgsqlParameter parameter)
   at Npgsql.NpgsqlParameter.ValidateAndGetLength()
   at Npgsql.NpgsqlParameterCollection.ValidateAndBind(ConnectorTypeMapper typeMapper)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,32 @@
 
 > The administration for the IdentityServer4 and Asp.Net Core Identity
 
-## Project Status
+## Big Thanks 🙏🏻 and an Update on the Project ⚡️
 
-[![Build status](https://ci.appveyor.com/api/projects/status/5yg59bn70399hn6s/branch/master?svg=true)](https://ci.appveyor.com/project/JanSkoruba/identityserver4-admin/branch/master)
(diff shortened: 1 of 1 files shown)


tuncaycemuzun (commented):
review successfull

WeihanLi:
related to #1030

